PR #103, opened 2014-06-29
Controller helper methods should have level access private.

DamirSvrtan:


andyw8:
I agree, this is a useful suggestion.

pirj:
This would be a nice addition, but it should at least mentioned that methods defined in modules/concerns that are included in the controller should also be private. Unless they define commonly used actions of course.

Anyway, closing as stale.

PR #116, opened 2014-08-05
use of reversible in migrations

akshaymohite:


syndbg:
Ping @akshaymohite

andyw8:
@akshaymohite I think this could be a useful addition, but it's been many years with no reply so I'm gong to close it. Feel free to re-open if you have time.

PR #124, opened 2014-10-16
Prefer symbol arguments to callbacks

mattbeedle:


cesc1989:
Agreed with this PR

andyw8:
Perhaps there could be be a configuration option of `MaxLinesInCallbackBody`, with a default of 5 or so.

mattbeedle:
Another reason why it's preferable is that you can then skip them with [`MyModel.skip_callback`](http://apidock.com/rails/ActiveSupport/Callbacks/ClassMethods/skip_callback) method. I don't think that's possible with the block style.

pirj:
Closing as stale.

PR #126, opened 2014-11-17
Fix typo describing foreign-key gems

tigershen23:


bbatsov:
This line was recently removed from the guide.

PR #132, opened 2015-02-21
Fix a typppppo

gsamokovarov:
I can imagine that those extra `ppppp` characters were not typed deliberately.

:beers:


bbatsov:
Wow, that's an epic typo! Thanks for spotting it!

PR #134, opened 2015-02-27
Syntax, language, consistency

AJFaraday:
First things first, loving the style guide, it was a really useful read. Thanks for putting this together.

This pull request is mostly a cople of style changes, inconsistent language, extra brackets etc.

There's two parts which have an editorial decision to be made...

* In defining scopes as a class method, I've added a note that these can not be added in the same way as named scopes...

```ruby
class User < ActiveRecord::Base

  def User.old
    where('age > ?', 80)
  end

  def User.heavy
    where('weight > ?', 200)
  end

end 
```

then in usage

```
> User.old.heavy
NoMethodError: undefined method `heavy' for #<ActiveRecord::Relation:0x0000000acfabe8>
```

* I've added a slightly clearer example for sexy validations, showing how you can convert multiple method calls to multiple arguments to the `validates` method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ programming resources.
   validates_length_of :email, maximum: 100 
  
   # good
-  validates :email, presence: true, length: {maximum: 100}
+  validates :email, presence: true, length: { maximum: 100 }
   ```
 
 * <a name="custom-validator-file"></a>
--- a/README.md
+++ b/README.md
@@ -423,6 +423,7 @@ programming resources.
   complicated, it is preferable to make a class method instead which serves the
   same purpose of the named scope and returns an `ActiveRecord::Relation`
   object. Arguably you can define even simpler scopes like this.
+  Note: this style of scoping can not be chained in the same way as named scopes.
 <sup>[[link](#named-scope-class)]</sup>
 
   ```Ruby
--- a/README.md
+++ b/README.md
@@ -587,7 +587,7 @@ when you need to retrieve a single record by some attributes.
   User.where(first_name: 'Bruce', last_name: 'Wayne').first
 
   # good
-  User.find_by(first_name: 'Bruce', last_name: 'Wayne'))
+  User.find_by(first_name: 'Bruce', last_name: 'Wayne')
   ```
 
 * <a name="find_each"></a>
--- a/README.md
+++ b/README.md
@@ -423,7 +423,7 @@ programming resources.
   complicated, it is preferable to make a class method instead which serves the
   same purpose of the named scope and returns an `ActiveRecord::Relation`
   object. Arguably you can define even simpler scopes like this.
-  Note: this style of scoping can not be chained in the same way as named scopes.
+  
 <sup>[[link](#named-scope-class)]</sup>
 
   ```Ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -367,9 +367,10 @@ programming resources.
   ```Ruby
   # bad
   validates_presence_of :email
-
+  validates_length_of :email, maximum: 100 
+ 
   # good
-  validates :email, presence: true
+  validates :email, presence: true, length: {maximum: 100}
```
bbatsov: Add spaces inside the hash.

README.md
```diff
@@ -422,6 +423,7 @@ programming resources.
   complicated, it is preferable to make a class method instead which serves the
   same purpose of the named scope and returns an `ActiveRecord::Relation`
   object. Arguably you can define even simpler scopes like this.
+  Note: this style of scoping can not be chained in the same way as named scopes.
```
bbatsov: I'd add this note after the code example with another code example.

bbatsov:
I see a few commits in this PR that can probably be removed (something was added and removed afterwards).

AJFaraday:
Yes, there was an area which was changed and then reverted, is this a problem if the finishing state is okay? What can I do to ammend this?

bbatsov:
@AJFaraday You can squash those commits with some other commits. Otherwise there're just noise in the history.

README.md
```diff
@@ -422,6 +423,7 @@ programming resources.
   complicated, it is preferable to make a class method instead which serves the
   same purpose of the named scope and returns an `ActiveRecord::Relation`
   object. Arguably you can define even simpler scopes like this.
+
```
bbatsov: This blank line is redundant.

README.md
```diff
@@ -431,6 +433,34 @@ programming resources.
     end
   end
   ```
+  
+  Note: this style of scoping can not be chained in the same way as named scopes. For instance:
```
bbatsov: You can rephrase this as "Note that...".

README.md
```diff
@@ -431,6 +433,34 @@ programming resources.
...
+      where('age > ?', 80)
+    end
+
+    def User.heavy
+      where('weight > ?', 200)
+    end
+
+  end 
+  ```
+  In this style both old and heavy work individually, but you can not call `User.old.heavy`, to chain these scopes use:
```
bbatsov: Add blank lines around this and put old and heavy in backticks.

README.md
```diff
@@ -431,6 +433,34 @@ programming resources.
     end
   end
   ```
+  
+  Note: this style of scoping can not be chained in the same way as named scopes. For instance:
+  
+  ```Ruby
+  # unchainable
+  class User < ActiveRecord::Base
+
```
bbatsov: Remove the blank lines around the class body.

README.md
```diff
@@ -431,6 +433,34 @@ programming resources.
...
+      where('weight > ?', 200)
+    end
+
+  end 
+  ```
+  In this style both old and heavy work individually, but you can not call `User.old.heavy`, to chain these scopes use:
+  ```Ruby
+  # chainable
+  class User < ActiveRecord::Base
+
```
bbatsov: Here as well.

AJFaraday:
I've made my first attempt here at a git rebase to remove the two extra commits, and now I'm not sure if I've done something wrong, but it looks like there's more commits.

AJFaraday:
Looking at the actual repo, it looks like the rebase worked, and the history on the pull request is showing them multiple times. I've made the requested changes.

bbatsov:
Something seems wrong to me - the commits exist several times with different committers. Maybe it'd be easier for you to squash them in a single commit with a longer description.

PR #138, opened 2015-04-03
Use the override setter instead of callbacks.

darkleaf:


seliverstov-maxim:
:+1:

bbatsov:
This is not bad advice, although one can argue this is not something Rails specific.

darkleaf:
It's OOP specific. The "valid?" method should only check model and return true or false. But in bad section it also change the model state. It's not OPP behavior.

With this approach in large projects it is very difficult to understand what is really going on.

For example, model may have several before_validation callbacks that modify one attribute.

http://en.wikipedia.org/wiki/Command%E2%80%93query_separation

darkleaf:
For example, we can see `has_secure_password` method.
https://github.com/rails/rails/blob/869a90512f36b04914d73cbf58317d953caea7c5/activemodel/lib/active_model/secure_password.rb#L105-L127

It define `password=`, that immediatly set `password_digest` without callbacks.

maicher:
I'd rather name this: "Don't modify object's state in before validation callbacks"

pirj:
Closing as stale.

PR #145, opened 2015-06-12
Prefer `Time` to `DateTime`

andyw8:
The linked gist is by a Rails core member, so I think it can be treated as being on good authority.


andyw8:
Sorry, this should be in the ruby style guide. I'll open a PR there instead.

PR #146, opened 2015-06-24
Fix erb comment format

Focinfi:


README.md
```diff
@@ -919,13 +919,13 @@ your application.
   methods don't.
 <sup>[[link](#url-not-path-in-email)]</sup>
 
-  ```Ruby
-  # bad
-  You can always find more info about this course
+  ```ERB
+  <%# bad %>
```
bbatsov: Add a space between % and #

bbatsov:
Please, change the commit message to `Fix erb comment format`.

KevinBongart:
Uh… what's the point of this PR? I'm not sure I understand

onebree:
@KevinBongart I think because since the code block would lie inside an ERB file, it is best for the guide to present that information. Also, it shows how to effectively make Ruby-style comments in ERB.

philoserf:
I do not believe this pull request fixes an error. Reading the existing example, I believe we have it right as is.

KevinBongart:
Yeah that's what I was thinking: while these changes are technically correct (I didn't know GitHub supported ERB syntax coloring), the rest of the file seem to be using standard Ruby comments

philoserf:
The #bad isn't really a comment. It is a market in the guide.

The line of text is not really a comment. It is content preceding the <%=.

bbatsov:
OK, guess we'll leave this as is, then. After all the meaning of the example is pretty clear even now.

PR #149, opened 2015-07-12
Add a note about shallow routes

freemanoid:
Add info about using shallow routes for nested routes more than 1 level deep.
This seems to be reasonable in order to avoid long urls for users (especially in case of long resource names like "payment_notification", "scheduled_appointment", etc.) and long url helpers for devs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,17 @@ programming resources.
     resources :comments
   end
   ```
+  
+* <a name="namespaced-routes"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
:+1:

onebree:
@freemanoid Can you expand this PR to include an example of what shallow routes look like? I never heard of the term before now.

freemanoid:
@onebree it's rails style guide so it's better for you to look into rails guides http://guides.rubyonrails.org/routing.html#nested-resources (section 2.7.2). Do you think that it's reasonable to include this link into the PR?

onebree:
Okay, so I have seen nested routes, just did not know about shallow routes. Thank you for sending me to the guide. I did not know it was included there!

PR #150, opened 2015-08-08
Add rails-assets. Add letter_opener.

juanpastas:


README.md
```diff
@@ -935,6 +935,14 @@ your application.
     # more settings
   }
   ```
+  
+* <a name="letter-opener"></a>
```
bbatsov: Newer version of Rails have built-in Emacs preview functionality, so I wouldn't recommend this.

bbatsov:
It's best to have unrelated changes in separate commits (and in this case - separate PRs).

README.md
juanpastas: That's great, where can I read about it?

bbatsov:
http://guides.rubyonrails.org/4_1_release_notes.html 2.4

juanpastas:
Thanks.

README.md
andyw8: I think @bbatsov meant to write Email rather than Emacs.

@juanpastas: [ActionMailer previews](http://guides.rubyonrails.org/4_1_release_notes.html#action-mailer-previews)

andyw8:
Thanks for the the suggestions @juanpastas but I think both of these are now outdated.

* Action Mailer Previews reduces the need for letter_opener.
* Webpacker etc. is now the standard approach now, so there's less need for https://rails-assets.org/

PR #152, opened 2015-08-09
Fixes #151

### LINKED ISSUE #151: Thoughts on the "Flawed gems" section
I think we should get rid of it. Instead of negative recommendations, let's add positive ones.

* A negative list is hard to mantain. A positive list too, but it's definitely easier to list the stuff one considers good, than everything one considers bad. For example, if there was a debugging section, one would say here "don't use `ruby-debug`, don't use `debugger` either" where one can just say "use `byebug`".
* A negative list is too subjective. For example, I consider `minitest-autotest` (`autotest` successor) to be as good as `guard`.
* A negative list doesn't have a good vibe. Let's focus on good stuff. Gems like `guard` or `simplecov` only get mentioned in this section, let's encourage their use instead of discouraging the alternatives.

deivid-rodriguez:
@katgironpe @bbatsov Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,6 @@ programming resources.
 * [Mailers](#mailers)
 * [Time](#time)
 * [Bundler](#bundler)
-* [Flawed Gems](#flawed-gems)
 * [Managing processes](#managing-processes)
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1096,30 +1096,10 @@ your application.
   same gem versions when they do a `bundle install`.
 <sup>[[link](#gemfile-lock)]</sup>
 
-## Flawed Gems
+## Awesome Gems
```
bbatsov: Just kill the section altogether as I have an aversion towards "awesome" lists.

bbatsov:
The commit message can be a bit more descriptive (and in sync with the project general commit message style).

katgironpe:
Agree with @bbatsov. Simply remove the section.

README.md
deivid-rodriguez: :+1:

deivid-rodriguez:
@bbatsov What's the general commit message style? I see some commit messages in present tense, some in past tense; some capitalized, some not; some ending with a dot, some not...

bbatsov:
Hmm, guess I didn't write a very extensive contributing.md. :-)

On all my projects I follow these rules http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

deivid-rodriguez:
Rebased it. Do you want me to rename the branch so the merge commit is clearer?

bbatsov:
That will do. Thanks!

deivid-rodriguez:
Thanks to you! :smiley:

PR #153, opened 2015-08-13
Clarified internationalization submodules paths

leehankyeol:
The original description is confusing a bit. To clarify it, I added `locales/` prior to original submodule paths such as `models` and `views`.


onebree:
:+1: At first, I thought it meant to place `locales` files in the folder `models`, not to create a subfolder. Good catch

ausmith:
:+1:

PR #154, opened 2015-08-31
Remove incorrect content about chaining scopes

Odaeus:
These methods return Relation objects that can be chained.


bbatsov:
:+1: I have no idea how this got in the guide in the first place. I guess I didn't read something carefully enough.

PR #157, opened 2015-10-15
Use the new hash syntax

rousisk:
Although this sample is under a "bad" code snippet, the thing to avoid is about something entirely different (i18n scoping), so I don't think we should use the older hash syntax anyway.


ausmith:
Sensible change to me.

PR #158, opened 2015-10-16
Added enum macro recommendation

Epigene:
Enums are a meaningful addition to Rails model ecosystem and can be used in the majority of models where a field can take a limited selection of values, but bool is insufficient.

Placing their definition under attr macros is one way to go about it, another might be to put them above attr macros, right under constants.
I'd be glad to hear others' opinion on the matter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,10 @@ programming resources.
     attr_accessor :formatted_date_of_birth
 
     attr_accessible :login, :first_name, :last_name, :email, :password
-
+    
+    # Rails4+ enums after attr macros, prefer the hash syntax
+    enum gender: { female: 0, male: 1 }
+    
     # followed by association macros
     belongs_to :country
 


onebree:
I agree that enums are important. I just learned about them this week, and feel as if not enough people talk about them.

KevinBongart:
I believe the hash syntax, while useful for customizing, brings a risk of breaking changes if the developers were to tweak the values (here, 0 and 1, but can get much bigger in some apps).

The alternative is simpler, straightforward, hides the configuration, but allows to be changed for customization if needed:

```ruby
enum gender: [:female, :male]
```

Epigene:
@KevinBongart I just do not see how the array syntax prevents "risk of breaking changes if the developers were to tweak the values".

In apps I have worked on there would be an enum with several options, for example

```ruby
enum status: [:new, :open, :stalled, :processing, :complete, :disabled]
``` 

I see two problems with this array syntax:
1) __Brittle__: chances of app breakage if symbol order is changed
2) __Lacks transparency__ which status `Model.pluck(:status).last #=> 4` refers to.

KevinBongart:
Good points :+1:

cesc1989:
IMO, as they represent an attribute in the model, they should go along the macros, however, enums have a different syntax, so placing them after the macros is ok.

PD:
Didn't know one could define enums using a hash. Although the ones I've written are pretty small, I can see (and read) how error prone is the array syntax.

Thanks for pointing it out, @Epigene

bbatsov:
Probably there should be a few rules about the usage of enums, as I feel many people would just miss this as part of a larger example.

bbatsov:
I'll merge this as-is and I hope someone will add some rules about enum usage down the road.

PR #162, opened 2015-11-11
Add the dependent option to the Active Record associations

maurogeorge:
Hi guys,

When we have a has many association without the `dependent` option like Chat
has many messages and we destroy the Chat, the Message objects still have the
`chat_id` reference pointing to a Chat that was destroyed.

```
[1] pry(main)> chat = Chat.first
                 :id => 1,
               :name => "Chat 01"
}
[2] pry(main)> message = chat.messages.first
            :id => 1,
       :message => "message 1",
       :chat_id => 1,
}
[3] pry(main)> chat.destroy
[4] pry(main)> message.reload
            :id => 1,
       :message => "message 1",
       :chat_id => 1,
}
[5] pry(main)> message.chat_id
1
[6] pry(main)> message.chat
nil
```

This patch add the `dependent` option to the has many and has one associations
to keep the things more explicit.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -553,6 +553,22 @@ programming resources.
   end
   ```
 
+* <a name="has_many-has_one-dependent-option"></a>
+  Define the `dependent` option to the `has_many` and `has_one` associations.
+<sup>[[link](#has_many-has_one-dependent-option)]</sup>
+
+  ```Ruby
+  # bad
+  class Post < ActiveRecord::Base
+    has_many :comments
+  end
+
+  # good
+  class Post < ActiveRecord::Base
+    has_many :comments, dependent: :destroy
+  end
+  ```
+
 ### ActiveRecord Queries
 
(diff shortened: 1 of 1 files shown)


ausmith:
Thoughts on using your `Chat` and `Message` analogy instead of the `User` analogy that you use in your change? I think it's clearer to explain a full `has_many` relationship.

maurogeorge:
Hi @ausmith thanks for the feedback, sorry for the delay :disappointed: 

I updated to use `Post` and `Comment` since we already have a example using these classes, makes sense?

ausmith:
Makes sense to me, but I'm not a committer on this project :smile:

bbatsov:
:+1:

bbatsov:
P.S. Additionally we can expand on the different `:dependent` options.

maurogeorge:
Thanks for the merge :heart:

PR #163, opened 2015-11-22
Add note about helper methods in models

alexander-lazarov:
Some notes about cluttering models with methods that should be helpers

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,14 @@ programming resources.
   For a more complete example refer to the
   [RailsCast on the subject](http://railscasts.com/episodes/326-activeattr).
 
+* <a name="model-business-logic"></a>
+  Unless they have some meaning in the business domain, don't put methods in
+  your model that just format your data (like code generating HTML). These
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -262,6 +262,14 @@ programming resources.
   For a more complete example refer to the
   [RailsCast on the subject](http://railscasts.com/episodes/326-activeattr).
 
+* <a name="model-business-logic"></a>
+  Unless they have some meaning in the business domain, don't put methods in
+  your model that just format your data (like code generating HTML). These
+  methods are most likely going to be called from the view layer only, so their
+  place is in helpers. Keep your models for business logic and data-persistance
```
bbatsov: Or even better - decorators. :-)

alexander-lazarov:
That being said maybe it's worth mentioning [draper](https://github.com/drapergem/draper)

bbatsov:
Yeah, we can mention it, although you don't really need to be using draper to be using presenters.

PR #167, opened 2015-12-25
Add ActiveSupport section

syndbg:
## Motivation:

Depend as little as possible on `ActiveSupport`.

## Change: 

Added tips about usage of `ActiveSupport` extensions:

* `try!`,
* `starts_with`,
* `ends_with`,
* `in?`,
* `forty_two`,
* `inquiry`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ programming resources.
 * [Internationalization](#internationalization)
 * [Assets](#assets)
 * [Mailers](#mailers)
+* [Active Support Core Extensions](#active-support-core-extensions)
 * [Time](#time)
 * [Bundler](#bundler)
 * [Managing processes](#managing-processes)
@@ -985,6 +986,22 @@ your application.
   sent. To overcome this emails can be sent in background process with the help
   of [sidekiq](https://github.com/mperham/sidekiq) gem.
 <sup>[[link](#background-email)]</sup>
+
+
+## Active Support Core Extensions
+
+* <a name="try-bang"></a>
+  Prefer Ruby 2.3.0's safe navigation operator `&.` over `ActiveSupport#try!`.
+<sup>[[link](#try-bang)]</sup>
+
+```ruby
+# bad
+obj.try! :fly
+
+# good
+obj&.fly
+```
+
 
 ## Time
 
--- a/README.md
+++ b/README.md
@@ -1002,6 +1002,21 @@ obj.try! :fly
 obj&.fly
 ```
 
+* <a name="active_support_aliases"></a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -986,6 +987,22 @@ your application.
   of [sidekiq](https://github.com/mperham/sidekiq) gem.
 <sup>[[link](#background-email)]</sup>
 
+
+## Active Support Core Extensions
+
+* <a name="try-bang"></a>
+  Prefer Ruby 2.3.0's safe navigation operator `&.` rather than `ActiveSupport#try!`.
```
bbatsov: rather than -> over

bbatsov:
I fill this section should be full of advice not to use those - especially the alias methods (e.g. `includes?`, `ends_with?`, etc).

README.md
syndbg: :+1:

syndbg:
@bbatsov I'll add them. Don't merge

syndbg:
@bbatsov Added more than I expected I would.

README.md
```diff
@@ -986,6 +987,91 @@ your application.
...
+```ruby
+# bad
+(1..50).to_a.forty_two
+
+# good
+(1..50).to_a[41]
+```
+
+* <a name="inquiry"></a>
+  Prefer Ruby's comparison operators over ActiveSupport's `Array#inquiry`, `Numeric#inquiry' and `String#inquiry`
```
ausmith: Ended up with a single quote instead of a backtick at the end of `Numeric#inquiry`

README.md
```diff
@@ -986,6 +987,91 @@ your application.
...
+
+# good
+ruby = 'two'
+ruby == 'two'
+
+# bad - Array#inquiry
+pets = %w(cat dog).inquiry
+pets.gopher?
+# good
+pets.include? 'cat'
```
ausmith: It feels self explanatory, but you didn't setup the `pets` object for your good example and it's still using the `inquiry` version.

ausmith:
Minus my nitpicks, I like this change. :+1:

syndbg:
Corrected.

syndbg:
@bbatsov 2.3 squiggly heredocs tip here -> https://github.com/bbatsov/ruby-style-guide/pull/523

The other comments are also addressed.


Btw, I was under the impression that markdown headers should have 2 lines between them and other content.

bbatsov:
> Btw, I was under the impression that markdown headers should have 2 lines between them and other content.

Nope. One line is just fine. I think this will work even without blank lines, but this is pretty hard to read.

PR #171, opened 2016-02-14
Fix typo

sachin21:
Fixed a typo: changed from **prefered** to **preferred**.

FIY: https://en.wiktionary.org/wiki/prefered


bbatsov:
:+1: Thanks!

PR #172, opened 2016-02-20
Transpose Examples in #sexy-validations

clay-reed-a:
The rule says to 'always use the new "sexy" validations', but the examples suggest "sexy" validations are bad.

We think "sexy" validations are good: they're easier to read.

The examples have been this way since inception w/ 60ca12c.

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,11 +442,11 @@ render plain: 'Ruby!'
 
   ```Ruby
   # bad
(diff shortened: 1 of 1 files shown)


bbatsov:
See https://rails.lighthouseapp.com/projects/8994/tickets/3058-patch-sexy-validations

PR #174, opened 2016-02-24
Update ruby formatter

vs4vijay:
Update ruby formatter

### DIFF
--- a/README.md
+++ b/README.md
@@ -987,7 +987,9 @@ your application.
   # bad
   You can always find more info about this course
   <%= link_to 'here', course_path(@course) %>
-
+  ```
+  
+  ```Ruby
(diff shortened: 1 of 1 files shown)


bbatsov:
Not sure what the point of this PR is.

vs4vijay:
Format Issue Image

@bbatsov Changed markdown, as it doesn't render properly. See screenshot for ref.

KevinBongart:
To me, this is more about function than style: the bad option just won't work, and the Rails docs already explain how to use these two methods correctly. @vs4vijay, do you agree with how I see the difference between this Rails style guide and the Rails docs?

andyw8:
This section was removed in #239.

PR #175, opened 2016-03-10
Remove duplicated advice for find_each method using.

harrykiselev:
Hello, Bozhidar!

Thanks for your gold mine of good advices.

Reading the style guides, I found advice that match twice ([one](https://github.com/bbatsov/rails-style-guide#find-each), [another one](https://github.com/bbatsov/rails-style-guide#find_each)) in the text (it seems to me):

So I found it necessary to remove the repetition.

Am I wrong? :)

Thanks.


bbatsov:
:+1: Thanks for spotting this!

PR #176, opened 2016-03-11
Remove unnecessary method call at good example of Array#inquiry section.

harrykiselev:
1) #inquiry method call in this example seems are unnecessary, but I may be wrong :)

2) Trim trailing white spaces from #158.


bbatsov:
:+1:

PR #178, opened 2016-04-01
Use symbols instead of numeric status codes

vbyno:
Rails has [symbols](http://billpatrianakos.me/blog/2013/10/13/list-of-rails-status-code-symbols/) for status codes. Why don't we use them by default? It would make more sense to the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,22 @@ render plain: 'Ruby!'
 ...
 ```
 
+* <a name="http-status-code-symbols"></a>
+  Prefer [corresponding symbols](https://gist.github.com/mlanett/a31c340b132ddefa9cca) to numeric HTTP status codes. They are meaningful and do not look like "magic" numbers for less known HTTP status codes.
+<sup>[[link](#http-status-code-symbols)]</sup>
+
+```Ruby
+# bad
+...
+render status: 500
+...
+
+# good
+...
+render status: :forbidden
+...
+```
+
 ## Models
 
 * <a name="model-classes"></a>


bbatsov:
Good suggestion. I've added a few tiny remarks, though.

README.md
```diff
@@ -280,6 +280,22 @@ render plain: 'Ruby!'
 ...
 ```
 
+* <a name="symbols-over-numeric-status-codes"></a>
+  Prefer corresponding symbols to numeric status codes. They are meaningful and do not look like "magic" numbers for less known status codes.
```
bbatsov: Use http status codes instead of just "status codes" for clarity's sake.

README.md
```diff
@@ -280,6 +280,22 @@ render plain: 'Ruby!'
...
+<sup>[[link](#symbols-over-numeric-status-codes)]</sup>
+
+```Ruby
+# not desired
```
bbatsov: Use "bad" here for consistency with the rest of the guide.

README.md
```diff
@@ -280,6 +280,22 @@ render plain: 'Ruby!'
...
+...
+render status: 500
+...
+
+# good
+...
+render status: :forbidden
```
bbatsov: Maybe we should also link to a place where someone can see all symbol codes.

vbyno: Should it be rails documentation? (http://guides.rubyonrails.org/layouts_and_rendering.html) this [gist](https://gist.github.com/mlanett/a31c340b132ddefa9cca) or the link I provided in merge request' description?

bbatsov: The gist looks great.

vbyno: Done! New suggestions are welcomed

dsounded:
Good point, because before this it was controversial.

MorganFujimaka:
Good job!

PR #180, opened 2016-05-14
Avoid `ActiveRecord::Base.suppress`

andyw8:
There is strong opposition to this in the Rails community:

https://www.reddit.com/r/ruby/comments/4j3097/rails_5_activerecord_suppress_a_step_too_far/
https://medium.com/spritle-software/rails-5-activerecord-suppress-a-step-too-far-d7ec2e4ed027#.lmh6aycc6
https://twitter.com/search?q=rails%20suppress

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,6 +648,18 @@ render status: :forbidden
     has_many :comments, dependent: :destroy
   end
   ```
+* <a name="avoid-suppress"></a>
+  Avoid `ActiveRecord::Base.suppress` (Rails 5).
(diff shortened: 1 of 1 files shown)


bbatsov:
I'm not sure whether it makes sense to add a rule for something unreleased just yet. Let's keep the conversation going here until Rails 5 is out.

andyw8:
As there has been no further comments in three years, I don't think is of much importance, so I'm going to withdraw my proposal.

PR #181, opened 2016-05-17
Never use instance variables in partials

killthekitten:
It is a bad practice to use instance variables in partials for several reasons. Some more of them can be found here http://rails-bestpractices.com/posts/2010/07/24/replace-instance-variable-with-local-variable/

### DIFF
--- a/README.md
+++ b/README.md
@@ -833,6 +833,27 @@ when you need to retrieve a single record by some attributes.
   Mitigate code duplication by using partial templates and layouts.
 <sup>[[link](#partials)]</sup>
 
+* <a name="no-instance-variables-in-partials"></a>
+  Avoid using instance variables in partials, pass a local variable to `render` instead. 
+  The partial may be used in a different controller or action, where the variable can have 
+  a different name or even be absent. In these cases, an undefined instance variable 
+  will not raise an exception whereas a local variable will.
+<sup>[[link](#no-instance-variables-in-partials)]</sup>
+
+```erb
+<!-- bad -->
+<!-- app/views/courses/show.html.erb -->
+<%= render 'course_description' %>
+<!-- app/views/courses/_course_description.html.erb -->
+<%= @course.description %>
+
+<!-- good -->
+<!-- app/views/courses/show.html.erb -->
+<%= render 'course_description', course: @course %>
+<!-- app/views/courses/_course_description.html.erb -->
+<%= course.description %>
+```
+
 ## Internationalization
 
 * <a name="locale-texts"></a>


README.md
```diff
@@ -833,6 +833,12 @@ when you need to retrieve a single record by some attributes.
   Mitigate code duplication by using partial templates and layouts.
 <sup>[[link](#partials)]</sup>
 
+* <a name="no-instance-variables-in-partials"></a>
+  Never use instance variables in partials, pass a local variable to `render` instead. Partial may be used
```
bbatsov: I'd change `never` to `avoid`, as plenty of people are doing so. Some concrete examples here would be helpful.

killthekitten: 👍 done

README.md
```diff
@@ -833,6 +833,26 @@ when you need to retrieve a single record by some attributes.
...
+  <%= render 'course_description' %>
+  # good
+  <%= render 'course_description', course: @course %>
+  
+  # app/views/courses/_course_description.html.erb
+  # bad
+  <%= @course.description %>
+  # good
+  <%= course.description %>
+  ```
```
bquorning: I think the meaning would be clearer if the entire bad example was in one listing, and the good one as well:
```erb
# bad
# app/views/courses/show.html.erb
<%= render 'course_description' %>
# app/views/courses/_course_description.html.erb
<%= @course.description %>

# good
# app/views/courses/show.html.erb
<%= render 'course_description', course: @course %>
# app/views/courses/_course_description.html.erb
<%= course.description %>
```

README.md
```diff
@@ -833,6 +833,26 @@ when you need to retrieve a single record by some attributes.
   Mitigate code duplication by using partial templates and layouts.
 <sup>[[link](#partials)]</sup>
 
+* <a name="no-instance-variables-in-partials"></a>
+  Avoid using instance variables in partials, pass a local variable to `render` instead. Partial may be used
```
andyw8: Can I suggest some small grammar changes:

**The** partial may be used in a different controller or action, where the variable can have a different name or **is absent**. In these cases **an** undefined instance variable will not raise an exception whereas **a** local variable will.

README.md
killthekitten: They say it's better late than never! Updated.

README.md
killthekitten: Yes, this way it reads so much better, thanks!

killthekitten:
I guess this one can be safely squashed back into master now?

bbatsov:
Indeed! Thanks!

PR #183, opened 2016-05-27
Use `find_by` instead of `find` to retrieve a single record by id

yamadayuki:
WHY
---

ref. https://github.com/bbatsov/rails-style-guide/issues/76#issuecomment-156909719
`ActiveRecord::FinderMethods#find` will raise `ActiveRecord::RecordNotFound` exception when we cannot retrieve the record by id. However, `ActiveRecord::FinderMethods#take` with `where` won't raise error if no record is found. So we should use `find_by` instead of `find`.
```rb
User.find(id)
# If no record is found, it raises ActiveRecord::RecordNotFound exception.

User.find_by(id: id)
# If no record is found, it returns nil.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -685,7 +685,7 @@ render status: :forbidden
   ```
 
 * <a name="find"></a>
-  Favor the use of `find` over `where`
+  Favor the use of `find_by` over `where`
 when you need to retrieve a single record by id.
 <sup>[[link](#find)]</sup>
 
@@ -694,7 +694,7 @@ when you need to retrieve a single record by id.
   User.where(id: id).take
 
   # good
-  User.find(id)
+  User.find_by(id: id)
   ```
 
 * <a name="find_by"></a>


KevinBongart:
Thank you for pointing out the subtle difference between `find` and `find_by`, but I'm not quite sure why this is a guideline. They're two different methods for a reason: each one serves a different use case.

1. If you need to raise an error when a record is not found, use `find(123)`
2. If you do not want an error when a record is not found, use `find_by(id: 123)`

Raising an error isn't necessarily a bad thing. It really depends on the context these methods are used in.

KevinBongart:
@yamadayuki Here's an example where the error-raising `find` is preferable:

```ruby
User.find_by(id: 123).email
# => NoMethodError: undefined method `email' for nil:NilClass
# Not a very useful error, can get difficult to debug.

User.find(123).email
# => ActiveRecord::RecordNotFound: Couldn't find User with 'id'={:id=>123}
# More much explicit error
```

In context:

```ruby
def password_reset_for_user(user_id)
  User.find(user_id).send_password_reset_instructions
rescue ActiveRecord::RecordNotFound
  # handle error, redirect to some useful information page
end
```

versus:

```ruby
def password_reset_for_user(user_id)
  User.find_by(id: user_id).send_password_reset_instructions
rescue NoMethodError
  # you don't know that the user wasn't found, for all we know maybe
  # someone deleted the "send_password_reset_instructions" method
end
```

yamadayuki:
@KevinBongart Thanks for your advice.
I'm sorry to persist in the compatibility with `where(id: id).take`. I think this change is negative and unnecessary for this guideline.

> 1. If you need to raise an error when a record is not found, use find(123)
2. If you do not want an error when a record is not found, use find_by(id: 123)

It is helpful for me! Thank you 👍 
I close this PR.

PR #190, opened 2016-07-11
When persisting AR objects always use the exception raising bang! method or handle the method return value. 

QuinnHarris:
Rule to go along with Rails/SaveBang cop https://github.com/bbatsov/rubocop/pull/3280

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,6 +649,38 @@ render status: :forbidden
   end
   ```
 
+* <a name="save-bang"></a>
+  When persisting AR objects always use the exception raising bang! method or handle the method return value. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -649,6 +649,38 @@ render status: :forbidden
   end
   ```
 
+* <a name="save-bang"></a>
+  When persisting AR objects always use the exception raising bang! method or handle the method return value. 
+  This applies to `create`, `save`, `update`, `destroy`, `first_or_create` and `find_or_create_by`
```
bbatsov: you forgot the final `.`

QuinnHarris: Thanks for catching all these.

bbatsov:
👍

PR #192, opened 2016-07-23
Add Rails 4 Test Prescriptions book reference

awolfson:
This book belongs in the "Further Reading" section as an equal of the other choices.


bbatsov:
:+1:

PR #194, opened 2016-08-05
Match good and bad examples for status code guideline

AlexVPopov:


AlexVPopov:
Closing in favor of #195, as 4XX codes are more likely to be manually returned than 5XX codes.

PR #195, opened 2016-08-05
Match good and bad examples for status code guideline

AlexVPopov:
When giving good/bad examples about the guideline for using status code symbols instead of numeric values, give a good and a bad example with the same status code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ render plain: 'Ruby!'
 ```Ruby
 # bad
 ...
-render status: 500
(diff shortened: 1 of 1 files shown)


AlexVPopov:
Ping :) Does this PR need any more work?

bbatsov:
Sorry about the super slow response! Too many projects, too little time! 

Need to find more editors!

AlexVPopov:
No problem ;)

PR #197, opened 2016-08-11
Favor Rails built-in ids method over pluck(:id)

AlexVPopov:


KevinBongart:
What's the reason behind this recommendation?

AlexVPopov:
Both behave and result in exactly the same thing, but `ids` is more expressive, and, in my opinion at least, stylistically preferable.

bf4:
On the other hand, it's a bigger change if you add another field to pluck...

AlexVPopov:
Do you encounter this situation that often? In most cases where I use `pluck(:id)`, I only need the ids, so that I can pass them to something, that specifically requires ids. If I need anything besides the ids, chances are, I'll need it right now (and use `pluck(:id, :some_other_attribute)`), than later.

bbatsov:
The changes look good, but your branch has to be rebased on top of the current `master` branch due to merge conflicts.

bbatsov:
Thanks!

PR #200, opened 2016-09-28
[Add] brasilian portuguese translation

abraaomiranda:


bbatsov:
Hmm. I see another similar PR - #201. Which translation is more complete?

PR #202, opened 2016-10-23
Add rule for environment dependent conditional variables

syndbg:
One way to reduce environment specific boilerplate.
In tests it's especially easy to stub an URL by accessing it with the `Rails.application.secrets.<key>`.

You might say, another way to handle this is by setting this in `config/environments/<environment>.rb` but it pollutes `Rails.application.config` and runs the risk of overriding already predefined Rails config keys and running into very _bullshit_ problems.

@bbatsov

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,53 @@ programming resources.
   Rails::Application.config_for(:yaml_file)
   ```
 
+* <a name="secrets-yaml"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
Seems odd to keep anything but access tokens here. I do understand it makes sense to some extent, but it hardly seems natural...

syndbg:
Closing since there's no activity here.

PR #207, opened 2016-11-23
Change column default value by more appropriate method

olegykz:
**change_column_default** method has self-descriptive name and it's usage is safer that change_column which is potentially danger (for example accidental column type change is possible).

One more advantage is that this method can be reversible in Rails 5 by specifying **from** & **to** parameter - [PR20018](https://github.com/rails/rails/pull/20018/files)


bbatsov:
👍

PR #209, opened 2016-12-25
Add rule of application file path  joining process

iguchi1124:
Related rubocop pull request: https://github.com/bbatsov/rubocop/pull/3822

Fix to use `Rails.root.join` clause when we join the rails application's file path.

@bbatsov

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,20 @@ programming resources.
   Rails::Application.config_for(:yaml_file)
   ```
 
+* <a name="rails-root-join"></a>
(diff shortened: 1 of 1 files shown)


KevinBongart:
To make this PR more constructive and the style guide more informative: why do you think are the bad practices bad? What makes the best practice better?

bbatsov:
@iguchi1124 ping :-)

andyw8:
@iguchi1124 Since it's been several years with no reply, I'm going to close this, but feel free to re-open if you have a chance.

PR #212, opened 2017-01-19
Small fix

marocchino:
wrong arguments.
http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attribute

PR #213, opened 2017-01-19
Fix typo

riseshia:
according to [api document](http://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attribute), it could get name, value, not hash :)


riseshia:
close this, dup #212

PR #219, opened 2017-09-11
Add recommendation to not order by `id`

Drenmi:
Recommends not doing things like:

```ruby
scope :chronological, -> { order(id: :asc) }
```

The sequence of ids is not guaranteed to be in any particular order, despite often (incidentally) being chronological. Use a timestamp column to order chronologically instead.

```ruby
scope :chronological, -> { order(created_at: :asc) }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -773,6 +773,22 @@ when you need to retrieve a single record by some attributes.
   # good
   User.where.not(id: id)
   ```
+
+* <a name ="order-by-id"></a>
+  Don't use the `id` column for ordering. The sequence of ids is not
+  guaranteed to be in any particular order, despite often (incidentally)
+  being chronological. Use a timestamp column to order chronologically.
+  As a bonus the intent is clearer.
+<sup>[[link](#order-by-id)]</sup>
+
+  ```ruby
+  # bad
+  scope :chronological, -> { order(id: :asc) }
+
+  # good
+  scope :chronological, -> { order(created_at: :asc) }
+  ```
+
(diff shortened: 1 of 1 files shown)


14113:
Maybe it's not related to `order` but what about to remain to add an index to column `created_at ` ?

Drenmi:
> Maybe it's not related to `order` but what about to remain to add an index to column `created_at` ?

An index should always be added on all columns that are used in a `where`. This is a more general rule. I'm not sure it should go in the style guide, since it's technically about database design. 🤔

Drenmi:
@bbatsov: You okay with this rule being added? I have a cop incoming. 😉

pocke:
> > Maybe it's not related to order but what about to remain to add an index to column created_at ?
>
> An index should always be added on all columns that are used in a where. This is a more general rule. I'm not sure it should go in the style guide, since it's technically about database design. 🤔


I think we should add a note for the index.
Because `id` is a primary key in almost cases. Searching by `id` is fast because always index for PK is added.
But maybe an index for `created_at` is not available. So, this change may cause performance issues.
When we apply this change, we need to be careful the performance issue.



> This is a more general rule. I'm not sure it should go in the style guide, since it's technically about database design.

Yes, it is a more general rule. But we have to be particularly careful in this case. So I think this guide should mention about the index.

bbatsov:
@Drenmi Hadn't reviewed the PRs here in a while. Sorry about that! :-) 

P.S. The note about the index is a valuable one.

Issue #208, opened 2016-12-01
Revisit Rails/DynamicFindBy


jcoyne:
https://github.com/bbatsov/rails-style-guide#find_by

What is the justification for this?  Isn't the dynamic method more expressive? Why should the style guide tell you not to use a feature of the framework? http://guides.rubyonrails.org/active_record_querying.html#dynamic-finders

olegykz:
http://edgeguides.rubyonrails.org/4_0_release_notes.html

> All dynamic methods except for find_by_... and find_by_...! are deprecated

jcoyne:
> except for

olegykz:
I'm not sure why they left find_by/find_by_...!, but AFAIK the main in issue with dynamic finders is that they based on **method_missing** ruby functionality with appropriate performance cost. IMHO, dynamic finders are less expressive and less readable that new finders API. But it's just IMHO :)

KevinBongart:
@olegykz That's a fair point, and I'm on board with @jcoyne that the current guideline doesn't make sense without this information. It reads like `find_by_x` is a bad idea, whereas it's a trade-off that should be left to the developer.

I'd rather see the guideline giving a reason:

```ruby
# bad
User.where(first_name: 'Bruce', last_name: 'Wayne').first

# good, readable but relies on Ruby's method_missing, thus coming with a performance cost
# see related deprecations: http://edgeguides.rubyonrails.org/4_0_release_notes.html#active-record-deprecations
User.find_by_first_name_and_last_name('Bruce', 'Wayne')

# preferable
User.find_by(first_name: 'Bruce', last_name: 'Wayne')
```

jcoyne:
@olegykz it's a one time performance cost (constant time), as they then define the method, which is then loaded in the method cache. This means there is no appreciable advantage in the time span for which a web-app would be up.  Clearly other people, by which I mean the Rails core team, feels differently about the usefulless of this functionality.

olegykz:
@jcoyne unfortunately I can't find the discussion about this functionality deprecation. Maybe another disadvantage of dynamic finders usage is that they doesn't work with associations or it potentially was too complex to implement support of them, so they decide to support less functionality amount with more universality and API consistency.

jcoyne:
[...]

Issue #76, opened 2013-10-14
where().first vs. find_by


kukula:
What do you think, what should I prefer in Rails: `where(field: value).first` or `find_by_field(value)` ?

rob-murray:
So `find_by_field(value)` and `where(field: value)` even with `where(field: value).last` are two different operations resulting in different SQL queries.

`find_by_field(value)` SQL has `LIMIT 1` applied so will be much more efficient than `where(field: value).last` - I would argue that these are not interchangeable and should be used in different circumstances.

If you expect a AR record to exist or not, ie binary decision, yes/no, use `find_by_field` otherwise if you expect many, irrespective of whether you want 1 or more, use `where(field: value)`.

bbatsov:
See https://github.com/bbatsov/rails-style-guide#find_by

kukula:
Great! Thank you @bbatsov !

xinspire:
The following are not equivalent:

```
# bad
User.where(id: id).take

# good
User.find(id)
```

If the id doesn't exist, the first one returns nil, while the second one throws a ActiveRecord::RecordNotFound exception.

fffx:
@xinspire And if id is a out of range value  `User.find(id)` will raise `ActiveRecord::RecordNotfound`, but `User.where(id: id).take` will raise `ActiveModel::RangeError`

see gitlab 500  https://gitlab.com/coldnight/ci-test/issues/464444444444444444444444444444444444

PR #220, opened 2017-10-18
Clarify find and find_by rules

wjordan:
This PR adds more specific description and examples to clarify the existing `find` and `find_by` rules.

Changes:

- Add an extra set of examples to favor `find_by` over `find_by_id` for the single-attribute case (`id: id`) in addition to the multiple-attribute case (`first_name: 'Bruce'` and `last_name: 'Wayne'`). Add comment clarifying the single-attribute dynamic finder is merely 'bad' style, while the multiple-attribute case is not only 'bad' but also 'deprecated in ActiveRecord 4.0, removed in 4.1+'. See #208 for related discussion.
- Clarify that `find` is favored only when you need to raise `ActiveRecord::RecordNotFound` when the record is not found, and compare it to the bang-methods `where.take!`, `find_by!`, and `find_by_id!` which exhibit similar behavior. See #183 and [#76 (comment)](https://github.com/bbatsov/rails-style-guide/issues/76#issuecomment-156909719) for related discussion.
- Use `where.take` instead of `where.first` in the `find_by` example, because `first` has different order-by behavior in ActiveRecord 4.0 and up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -734,31 +734,43 @@ render status: :forbidden
   ```
 
 * <a name="find"></a>
-  Favor the use of `find` over `where`
-when you need to retrieve a single record by id.
+  Favor the use of `find` over `where.take!`, `find_by!`, and `find_by_id!`
+  when you need to retrieve a single record by primary key id and raise
+  `ActiveRecord::RecordNotFound` when the record is not found.
 <sup>[[link](#find)]</sup>
 
   ```ruby
   # bad
-  User.where(id: id).take
+  User.where(id: id).take!
 
+  # bad
+  User.find_by_id!(id)
+
+  # bad
+  User.find_by!(id: id)
+  
   # good
   User.find(id)
   ```
 
 * <a name="find_by"></a>
-  Favor the use of `find_by` over `where` and `find_by_attribute`
-when you need to retrieve a single record by some attributes.
+  Favor the use of `find_by` over `where.take` and `find_by_attribute`
+  when you need to retrieve a single record by one or more attributes and return
+  `nil` when the record is not found.
 <sup>[[link](#find_by)]</sup>
 
   ```ruby
   # bad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -734,31 +734,43 @@ render status: :forbidden
   ```
 
 * <a name="find"></a>
-  Favor the use of `find` over `where`
-when you need to retrieve a single record by id.
+  Favor the use of `find` over `where.take!`, `find_by!`, and `find_by_id!`
```
KevinBongart: This reads like `find_by_id!` is a recommended method whereas it is listed as bad a few lines below. Can you clarify the copy a little?

wjordan: > This reads like `find_by_id!` is a recommended method whereas it is listed as bad a few lines below. Can you clarify the copy a little?

"Favor the use of A over B, C, and D" reads clearly (at least to me) that A is the only favored item, while B, C, and D are not, so I'm a little confused by your comment- what exactly are you suggesting needs to be clarified?

KevinBongart: But isn’t `find_by!` recommended over `find_by_id!`?

wjordan: No- when you need to retrieve a record by primary key id and raise `ActiveRecord::NotFound` when the record is not found, `find` is favored over _any_ of the bang-methods. So `find_by!` is not recommended over `find_by_id!` for this case because neither is recommended at all.

I still can't think of any clearer way to state this other than "Favor the use of `find` over `where.take!`, `find_by!`, and `find_by_id!`, when you need to retrieve a single record by primary key id and raise `ActiveRecord::RecordNotFound` when the record is not found." Feel free to offer an alternative if you have one.

KevinBongart: Got it. Sorry I misread, I initially thought you were discouraging the use of `find_by!`, but in this case you're specifically talking about primary key queries. 👍

bbatsov:
👍

PR #224, opened 2018-03-21
Add size vs count/length rule

swrobel:
Source: http://batsov.com/articles/2014/02/17/the-elements-of-style-in-ruby-number-13-length-vs-size-vs-count/#comment-1248998887

### DIFF
--- a/README.md
+++ b/README.md
@@ -819,6 +819,21 @@ when you need to retrieve a single record by some attributes.
   SELECT\n    users.id, accounts.plan\n  FROM\n    users\n  INNER JOIN\n    acounts\n  ON\n    accounts.user_id = users.id
   ```
 
+* <a name="size-over-count-or-length"></a>
+  When querying ActiveRecord collections, prefer `size` (selects between count/length behavior based on whether collection is already loaded) or `length` (always loads the whole collection and counts the array elements) over `count` (always does a database query for the count).
+<sup>[[link](#size-over-count-or-length)]</sup>
+
+  ```ruby
+  # bad
+  User.count
+
+  # good
+  User.all.size
+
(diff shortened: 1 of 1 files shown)


bbatsov:
I think some important explanations from the original comment are missing here:

```There's also the differences between `#count`, `#length`, and `#size` on ActiveRecord::Relation. `#count` will always perform a count query, `#length` will load up the records, and call `#length` on the resulting array, and `#size` will do the smart thing based on whether or not we've loaded. TLDR: use `#size`, unless you're calling it immediately before you need to load the real records, in which case use `#length`. Never use `#count`, since `#size` will skip the query if it can.

```

swrobel:
@bbatsov good point, I clearly didn't fully grok the comment! Updated to reflect that size/length is preferred over count.

README.md
```diff
@@ -819,6 +819,22 @@ when you need to retrieve a single record by some attributes.
   SELECT\n    users.id, accounts.plan\n  FROM\n    users\n  INNER JOIN\n    acounts\n  ON\n    accounts.user_id = users.id
   ```
 
+* <a name="size-over-count-or-length"></a>
+  When querying ActiveRecord collections, prefer `size` (selects between count/length behavior based on whether collection is already loaded) or `length` (always loads the whole collection and counts the array elements) or `count` (always does a database query for the count).
```
bbatsov: I guess the final `or` should be `over` right?

zhisme:
Hey there!
Is there any reason to load all users in too memory with `#length` method.
When do people use it? I can't find any real-case example.

bbatsov:
@zhisme You might want to prevent n+1 queries afterwards if you know you'll be doing some processing of all those user records.

PR #226, opened 2018-04-04
Remove Numeric#inquiry as bad

bdewater:
This has been upstreamed in Ruby 2.3: https://ruby-doc.org/core-2.3.0/Numeric.html#method-i-negative-3F

It also allows writing `(-2..2).select(&:positive?)` instead of `(-2..2).select { |i| i > 0 }` which is in line with https://github.com/bbatsov/ruby-style-guide#single-action-blocks

### DIFF
--- a/README.md
+++ b/README.md
@@ -1303,7 +1303,7 @@ obj&.fly
 ```
 
 * <a name="inquiry"></a>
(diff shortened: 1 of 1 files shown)


bbatsov:
👍

PR #227, opened 2018-04-17
single attribute validations

mreinsch:
avoid having multiple attributes listed in one `validates` expression which makes it harder to read

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,21 @@ render status: :forbidden
   validates :email, presence: true, length: { maximum: 100 }
   ```
 
+* <a name="single-attribute-validations"></a>
+  To make validations easy to read, don't list multiple attributes per
+  validation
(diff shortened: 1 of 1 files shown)


KevinBongart:
Would it make sense to only consider it bad if there's more than one validation per attribute?

There's nothing wrong with just this:
```ruby
validates :email, :password, presence: true
```

mreinsch:
@KevinBongart I'd argue to generally avoid using more than one attribute, even when using

```
validates :email, :password, presence: true
```

First, when reading models I find it very easy to miss that this validation is for both email and password even when it's standalone. Beginners can also quite easily confuse this with `validates :email, password: true, presence: true`. 
Second, when more validations are added more changes are required to make this consistent with the style guide.

bbatsov:
Thanks!

PR #228, opened 2018-05-29
Add guideline for ordering of ActiveRecord callback declarations

AlexVPopov:


bbatsov:
Good suggestion! That would also be a fine cop!

Issue #233, opened 2018-09-03
Don't use .to_time


RST-J:
The section about [time](https://github.com/rubocop-hq/rails-style-guide#time) suggests not to use `Time.parse`. Rails core-extension `.to_time` for the `String` class also returns `Time`, so I think the use of `.to_time` should also be discouraged.

KevinBongart:
> Rails core-extension `.to_time` for the `String` class also returns `Time`

KevinBongart:
> Rails core-extension `.to_time` for the `String` class also returns `Time`

How does this discourage against using `String#to_time`? On the contrary, it seems like Rails supports it

RST-J:
Ah, I see. The point is not primarily the return type but how the input is interpreted. I mistakenly thought that belongs together.

However, the given reason not to use `Time.parse` is that it interprets the given time in the system's time zone.
```ruby
# bad => Will assume time string given is in the system's time zone.
Time.parse('2015-03-02 19:05:37')
```
And `String.to_time` does exactly the same by default (or when given `:local` as parameter). It does not obey the configured time zone.

pirj:
This makes sense
```
# bad
Time.parse("13-12-2012") # => 2012-12-13 00:00:00 +0400
# bad as well 
"13-12-2012".to_time # => 2012-12-13 00:00:00 +0400
# good
Time.zone.parse("13-12-2012") # => Thu, 13 Dec 2012 00:00:00 EST -05:00
```

Will you please send a pull request?

PR #234, opened 2018-09-18
Add section about String#to_time

RST-J:
As I reported in #233, `String#to_time` has the same behaviour as `Time.parse` and hence probably should also be considered bad style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1463,8 +1463,8 @@ pets.include? 'cat'
 <sup>[[link](#to-time)]</sup>
 
   ```ruby
-  # bad
-  '2015-03-02 19:05:37'.to_time # => Will also assume time string given is in the system's time zone.
+  # bad - assumes time string given is in the system's time zone.
+  '2015-03-02 19:05:37'.to_time
 
   # good
   Time.zone.parse('2015-03-02 19:05:37') # => Mon, 02 Mar 2015 19:05:37 EET +02:00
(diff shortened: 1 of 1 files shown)


RST-J:
rubocop-rails already supports [this](https://github.com/rubocop-hq/rubocop-rails/blob/master/lib/rubocop/cop/rails/date.rb#L38).

README.md
```diff
@@ -1458,6 +1458,18 @@ pets.include? 'cat'
   Time.zone.parse('2015-03-02 19:05:37') # => Mon, 02 Mar 2015 19:05:37 EET +02:00
   ```
 
+* <a name="to-time"></a>
+  Don't use [`String#to_time`](https://apidock.com/rails/String/to_time)
+<sup>[[link](#to-time)]</sup>
+
+  ```ruby
+  # bad
+  '2015-03-02 19:05:37'.to_time # => Will also assume time string given is in the system's time zone.
```
pirj: `=>` construct implies what follows it is returned by the code to the left, and this might be confusing. WDYT of:
```
# bad - assumes time string given is in the system's time zone
'2015-03-02 19:05:37'.to_time

# good - is using time zone from application configuration
  Time.zone.parse('2015-03-02 19:05:37') # => Mon ...
```

I understand that in the same paragraph the confusing syntax is used, probably to be fixed in a follow-up.

bbatsov:
👍

PR #235, opened 2018-09-20
update markdown indention

3limin4t0r:
Changed indention from 2 to 4 spaces.

Requested in https://github.com/rubocop-hq/ruby-style-guide/pull/729#issuecomment-423189354.

It's worth pointing out that with the extra indention some lines pass the column 80 mark, however this was already the case for some lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,14 @@ You can generate a PDF or an HTML copy of this guide using
 
 Translations of the guide are available in the following languages:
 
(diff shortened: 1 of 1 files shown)


3limin4t0r:
With the resulting [pdf].

[pdf]: ../files/2401971/rails-style-guide.pdf "rails-style-guide.pdf"

PR #236, opened 2019-02-19
Use "find(id)" instead of "find_by(id: id)"

iago-silva:
"find_by(id: id)" was used as good practice

### DIFF
--- a/README.md
+++ b/README.md
@@ -843,7 +843,7 @@ render status: :forbidden
     User.find_by_first_name_and_last_name('Bruce', 'Wayne')
 
     # good
-    User.find(id)
+    User.find_by(email: email)
     User.find_by(first_name: 'Bruce', last_name: 'Wayne')
     ```
 
--- a/README.md
+++ b/README.md
@@ -834,11 +834,11 @@ render status: :forbidden
 
     ```ruby
     # bad
-    User.where(id: id).take
+    User.where(email: email).take
     User.where(first_name: 'Bruce', last_name: 'Wayne').take
 
     # bad
-    User.find_by_id(id)
+    User.find_by_email(email)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -843,7 +843,7 @@ render status: :forbidden
     User.find_by_first_name_and_last_name('Bruce', 'Wayne')
```
pirj: Do you think it makes sense to update the desction to "Favor the use of `find_by`/`find`"?

eliotsykes:
Take care of the different behaviour these have.

`User.find(id)` will raise an error if there is no user with that id. `User.find_by(id: id)` will return `nil`.

To avoid confusion, it may be better to remove the `id` examples entirely and use a different attribute that isn't the primary key, e.g. `email`.

st0ck:
I believe we can't just replace `find_by(id: id)` by `find(id)`. These methods have different results.
`find(id)` will fail if there is no such id, while `find_by(id: id)` will return `nil` in this case.

README.md
iago-silva: It makes sense, but the below suggestion seems be more careful 🤠.

iago-silva:
> Take care of the different behaviour these have.
> 
> `User.find(id)` will raise an error if there is no user with that id. `User.find_by(id: id)` will return `nil`.
> 
> To avoid confusion, it may be better to remove the `id` examples entirely and use a different attribute that isn't the primary key, e.g. `email`.

README.md
```diff
@@ -843,7 +843,7 @@ render status: :forbidden
...
 
     # good
-    User.find(id)
+    User.find_by(email: email)
```
eliotsykes: The 2 corresponding `#bad` examples above should also use `email` instead of `id`.

iago-silva:
Do I need do more something to merge it?

PR #237, opened 2019-02-28
Change language for validation styles

andyw8:
This matches the wording used in the `Rails/Validation` cop.


bbatsov:
Probably we can come up with better naming, but I guess that will do. Thanks!

PR #239, opened 2019-04-22
Remove section about mailer paths

andyw8:
I think this advice doesn't belong in the style guide.

Using `_path` instead of `_url` is simply a bug, it's not a matter of style or preference.


bbatsov:
Good point!

PR #244, opened 2019-04-25
Update advice for where formatting code should live

andyw8:
I think it would good to point out that while helpers may be good for simple formatting, when the formatting becomes more complex it can be a problem.

There's a wide variety of alternative approaches (decorators, presenters, Cells) so I don't want to be overly-specific here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,9 +355,9 @@ render status: :forbidden
     Unless they have some meaning in the business domain, don't put methods in
     your model that just format your data (like code generating HTML). These
     methods are most likely going to be called from the view layer only, so their
-    place is in helpers, if simple, or a separate class if more complex.
-    Keep your models for business logic and data-persistence
-    only.
+    place is in helpers, if simple, or a separate Decorator or Presenter class
+    if more complex.
+    Keep your models for business logic and data-persistence only.
     <sup>[[link](#model-business-logic)]</sup>
     
     ```erb
--- a/README.md
+++ b/README.md
@@ -360,26 +360,48 @@ render status: :forbidden
     only.
     <sup>[[link](#model-business-logic)]</sup>
     
+    ```erb
+    <!-- bad -->
+    <% if widget.price == 0 %>
+       Free
+    <% else %>
+       widget.price * customer.discount %>
+    <% end %>
+    ```
+    
+    ```erb
+    <!-- better -->
+    <%= catalog_price(widget, customer) %>
+    <% end %>
+    ```
+    
     ```ruby
-    # bad
+    # better
     module WidgetsHelper
-      def catalog_price(widget)
-        # (lots of complex code)
+      def catalog_price(widget, customer)
+        if widget.price == 0 
+          "Free"
+        else
+          widget.price * customer.discount
+        end
       end
     end
     ```
     
     ```ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -355,7 +355,8 @@ render status: :forbidden
     Unless they have some meaning in the business domain, don't put methods in
     your model that just format your data (like code generating HTML). These
     methods are most likely going to be called from the view layer only, so their
-    place is in helpers. Keep your models for business logic and data-persistence
+    place is in helpers, if simple, or a seperate class if more complex.
```
eliotsykes: - Typo on separate
- Consider illustrating "separate class" with related pattern names to assist readers on how to proceed and what to look up next

> or a separate class (e.g. presenter, value object) if more complex

bbatsov:
Might be nice to add some simple examples, though. I've noticed that people tend to understand a lot better if there's some code to go with them.

README.md
```diff
@@ -355,9 +355,56 @@ render status: :forbidden
...
+    Keep your models for business logic and data-persistence
     only.
     <sup>[[link](#model-business-logic)]</sup>
+    
+    ```erb
+    <!-- bad (if complex) -->
+    <% if widget.price == 0 %>
+       Free
+    <% else %>
+       widget.price * customer.discount %>
```
bbatsov: The opening `<%` is missing.`

andyw8: Thanks, I've fixed.

andyw8:
This isn't ideal because the calculation part should probably be considered business logic. I'll try come up with a better example.

README.md
```diff
@@ -355,9 +355,55 @@ render status: :forbidden
...
+    <% if widget.price == 0 %>
+       Free
+    <% else %>
+       <%= widget.price * customer.discount %>
+    <% end %>
+    ```
+    
+    ```erb
+    <!-- better -->
+    <%= catalog_price(widget, customer) %>
```
pirj: How do you feel about `widget.price_for(customer)`?

README.md
```diff
@@ -355,9 +355,55 @@ render status: :forbidden
...
     your model that just format your data (like code generating HTML). These
     methods are most likely going to be called from the view layer only, so their
-    place is in helpers. Keep your models for business logic and data-persistence
+    place is in helpers, if simple, or a separate class if more complex.
+    Keep your models for business logic and data-persistence
     only.
     <sup>[[link](#model-business-logic)]</sup>
+    
+    ```erb
+    <!-- bad (if complex) -->
```
pirj: An example of non-business logic in a model can be presentation of the price depending on the current user's currency settings, user name/phone number presentation,

README.md
andyw8: I was aiming to indicate that a helper is a 'half-way' solution between having the view logic in the template and introducing a separate object. Adding a `price_for` method to `widget` would would mean the view logic to the model, which although might appear 'cleaner', is not something I want to encourage.

README.md
andyw8: Spelling fixed in 663f1f28922be10daf3cbd82bf9205bb6386a6ed.

I've made the guidelines more specific in bf43762.

andyw8:
Closing until I can pick a better example.

PR #246, opened 2019-04-27
Update links to HTTPS

andyw8:
Update links to HTTPS (where possible)

PR #248, opened 2019-05-02
Replace apidock.com links with official docs

andyw8:
As mentioned by @pirj in #246, since apidock is no longer updated.


bbatsov:
👍

PR #250, opened 2019-06-01
Convert the guide to AsciiDoc

pirj:
# Description

AsciiDoc allows us to have native admonitions, footnotes, sane nesting of list items, etc and to easily export the guide in various formats.
That would also make it trivial to publish a simple site from the guide down the road if we want to.

A longer term plan is to convert all guides to AsciiDoc and maybe even RuboCop's own documentation.
RSpec style guide [has been recently converted](https://github.com/rubocop-hq/rspec-style-guide/pull/97).

Appreciate if you proofread or even just glance over https://github.com/rubocop-hq/rails-style-guide/blob/convert-to-asciidoc/README.adoc


bbatsov:
Everything seems good to me in the converted documented. Great work, @pirj! 🙇 Two down, just one to go!

PR #252, opened 2019-06-09
Use proper Title Case in section titles

pirj:
Section titles were mostly auto-generated during AsciiDoc conversion, and some of them were off.
But the main culprit that they were not using proper Title Case.


koic:
Thanks!

PR #253, opened 2019-06-28
Use proper names of Rails compornents

koic:
This PR uses proper names of Rails compornents.

> The proper names of Rails components have a space in between the words,
> like “Active Support”. `ActiveRecord` is a Ruby module, whereas Active Record is an ORM.

https://github.com/rails/rails/blob/v6.0.0.rc1/guides/source/api_documentation_guidelines.md#wording

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -164,7 +164,7 @@ end
 
 === Nested Routes [[nested-routes]]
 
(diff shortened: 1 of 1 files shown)


bbatsov:
Nice!

PR #255, opened 2019-07-08
Add section for hash syntax of enums

tejasbubane:
Closes https://github.com/rubocop-hq/rails-style-guide/issues/144

Ref: https://github.com/rubocop-hq/rubocop-rails/issues/78

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -385,6 +385,26 @@ class Transaction < ActiveRecord::Base
 end
 ----
 
+=== Enums [[enums]]
+
+Prefer using the hash syntax for `enum`. Array makes the database values implicit
+& any insertion/removal/rearrangement of values in the middle will most probably
+lead to broken code.
+
+[source,ruby]
+----
+class Transaction < ActiveRecord::Base
+  # bad - implicit values - ordering matters
+  enum type: %i[credit debit]
+
+  # good - explicit values - ordering does not matter
+  enum type: {
+    credit: 0,
+    debit: 1
+  }
+end
+----
+
 === Macro Style Methods [[macro-style-methods]]
 
 Group macro-style methods (`has_many`, `validates`, etc) in the beginning of the class definition.
@@ -403,7 +423,7 @@ class User < ActiveRecord::Base
 
   attr_accessible :login, :first_name, :last_name, :email, :password
 
-  # Rails 4+ enums after attr macros, prefer the hash syntax
+  # Rails 4+ enums after attr macros
   enum gender: { female: 0, male: 1 }
 
   # followed by association macros


README.adoc
```diff
@@ -385,6 +385,24 @@ class Transaction < ActiveRecord::Base
 end
 ----
 
+=== Enums [[enums]]
+
+Prefer hash syntax over arrays for `enums`.
```
pirj: WDYT of
```
`enum`s
```
or 
Prefer using the hash syntax for `enum`.

README.adoc
```diff
@@ -385,6 +385,24 @@ class Transaction < ActiveRecord::Base
 end
 ----
 
+=== Enums [[enums]]
```
pirj: There's this example:
```ruby
# Rails 4+ enums after attr macros, prefer the hash syntax
```

Do you think it becomes redundant with this new guideline?

README.adoc
```diff
@@ -385,6 +385,24 @@ class Transaction < ActiveRecord::Base
...
+[source,ruby]
+----
+class Transaction < ActiveRecord::Base
+  # bad - implicit values - ordering in array matters
+  enum type: %i[credit debit]
+
+  # good - explicit values - ordering does not matter
+  enum type: {
+    credit: 0,
+    debit: 1,
```
pirj: Leading comma.

README.adoc
```diff
@@ -385,6 +385,24 @@ class Transaction < ActiveRecord::Base
...
 ----
 
+=== Enums [[enums]]
+
+Prefer hash syntax over arrays for `enums`.
+
+[source,ruby]
+----
+class Transaction < ActiveRecord::Base
+  # bad - implicit values - ordering in array matters
```
pirj: Can't come up with better wording, do you think that "ordering" is sufficient to emphasize that if something is inserted or removed in the middle, things will break, not only when entries are rearranged?

README.adoc
tejasbubane: Yes, I saw the guideline exists (hence placed this rule just above). I feel the comment reiterates the rule & don't see much of an issue. I can remove `, prefer the hash syntax` if required.

README.adoc
tejasbubane: Changed to `Prefer using the hash syntax for enum.`

README.adoc
tejasbubane: Removed.

README.adoc
tejasbubane: Removed.

README.adoc
tejasbubane: Added more details to the rule description.

README.adoc
```diff
@@ -385,6 +385,26 @@ class Transaction < ActiveRecord::Base
 end
 ----
 
+=== Enums [[enums]]
```
koic: I investigated whether `=== Enums` or `=== Enum` would be better. The API says `Enum`, but the guide says `Enums`, so I think that `Enums` would be fine.

- Guide ... https://guides.rubyonrails.org/active_record_querying.html#enums
- API ... https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

PR #256, opened 2019-08-21
Use enum example with User#preferred_indentation

olleolleolle:


README.adoc
```diff
@@ -424,7 +424,7 @@ class User < ActiveRecord::Base
   attr_accessible :login, :first_name, :last_name, :email, :password
 
   # Rails 4+ enums after attr macros
-  enum gender: { female: 0, male: 1 }
+  enum preferred_indentation: { tabs: 0, spaces: 1 }
```
mroderick:
```suggestion
  enum preferred_indentation: { tabs: 0, two_spaces: 1, four_spaces: 2 }
```

koic:
Could you describe the intention of this change in the PR description (and commit message) ?

dhnaranjo:
@koic, I imagine @olleolleolle's reasoning is the same as mine: An enum attribute is one that has a specific set of valid values that isn't a true/false question, a number, or a what have you. The current example, gender, with only male and female values, is making an assertion that these are the only two valid gender expressions. Many people do not identify with these options, and it's a pretty emotional topic.

There's no benefit for purpose of the example to potentially hurt or exclude readers of this guide. The alternative proposed in this PR still communicates the intent of the positioning rule without that risk.

bbatsov:
I'm totally fine with changing the existing example, but I feel that the suggested alternative is not much better - it's not like we don't have a million indentation options available. :-) Maybe something with a status of something? Or days of week? Let's just make sure it's something easy to understand and clear cut.

koic:
@dhnaranjo I get it. Thank you for the explanation. And I agree with @bbatsov.
I'd like to reopen and merge https://github.com/rubocop-hq/rails-style-guide/pull/257 because `role` would be appropriate for an example.

olleolleolle:
@koic Good resolution, I'll close this issue!

PR #257, opened 2019-10-26
Changed example enum

dhnaranjo:
I think gender isn't a good example for a specifically limited set of valid options. Changing it to a user role example both communicates the intended message and isn't exclusionary.


dhnaranjo:
I noticed there was another PR addressing the same issue, so I'll discuss it there.

koic:
Thanks!

PR #259, opened 2020-01-22
Fix typo at Controller Rendering chapter

hungnh103:
A minor typo that I accidentally encounter when referencing this style guide


koic:
Thanks!

PR #265, opened 2020-07-01
Remove an incorrect comment for `find_by` example

koic:
Dynamic finder methods (`DynamicMatchers`)  has not been previously deprecated.
AFAIK, Rails 6.0 has also no plans to deprecate it.
https://github.com/rails/rails/blob/v6.0.3.2/activerecord/lib/active_record/dynamic_matchers.rb

So, it seems that this was written incorrectly when I refer to the following.

- https://github.com/rubocop-hq/rails-style-guide/pull/220
- https://github.com/rubocop-hq/rails-style-guide/issues/208

Therefore this PR removes the incorrect example comment for `find_by`.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -831,7 +831,6 @@ User.where(first_name: 'Bruce', last_name: 'Wayne').take
 
 # bad
 User.find_by_email(email)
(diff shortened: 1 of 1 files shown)


pirj:
Nice catch, thanks!

PR #270, opened 2020-09-21
Extend query conditions guideline to cover equality

eugeneius:
Passing conditions as a hash is always preferable to using an SQL fragment, regardless of whether the conditions are negated.


pirj (approved):
Thank you!

koic:
Thanks!

PR #275, opened 2020-11-24
Add three-state boolean problem

tejasbubane:
Issue: https://github.com/rubocop-hq/rubocop-rails/issues/337

Ref: https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1026,6 +1026,25 @@ end
 While enforcing table defaults only in Rails is suggested by many Rails developers, it's an extremely brittle approach that leaves your data vulnerable to many application bugs.
 And you'll have to consider the fact that most non-trivial apps share a database with other applications, so imposing data integrity from the Rails app is impossible.
 
+=== 3-state Boolean [[three-state-boolean]]
+
+With SQL databases, if a boolean column is not given a default value, it will have three possible values: `true`, `false` and `NULL`.
+Boolean operators [work in unexpected ways](https://en.wikipedia.org/wiki/Three-valued_logic) with `NULL`.
+
+For example in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
+
+To avoid such situations, boolean columns should always have a default value and a `NOT NULL` constraint.
+
+[source,ruby]
+----
+# bad - boolean without a default value
+add_column :users, :active, :boolean
+
+# good - boolean with a default value (`false` or `true`) and with restricted `NULL`
+add_column :users, :active, :boolean, default: true, null: false
+add_column :users, :admin, :boolean, default: false, null: false
+----
+
 === Foreign Key Constraints [[foreign-key-constraints]]
 
 Enforce foreign-key constraints. As of Rails 4.2, Active Record supports foreign key constraints natively.


README.adoc
```diff
@@ -1014,6 +1014,20 @@ end
 While enforcing table defaults only in Rails is suggested by many Rails developers, it's an extremely brittle approach that leaves your data vulnerable to many application bugs.
 And you'll have to consider the fact that most non-trivial apps share a database with other applications, so imposing data integrity from the Rails app is impossible.
 
+=== 3-state Boolean [[three-state-boolean]]
+
+When boolean columns are not given default value, they end up having three possible values: `true`, `false` and `NULL`.
```
bbatsov, line 1019: I guess that depends on the database, so you might want to expand on this.

README.adoc
```diff
@@ -1014,6 +1014,20 @@ end
...
+Boolean operators work in unexpected ways with `NULL`. Avoid such situations by always adding default values to booleans columns.
```
bbatsov, line 1020: Without examples it's hard to tell what you mean by this.

README.adoc
```diff
@@ -1014,6 +1014,20 @@ end
...
+
+[source,ruby]
+----
+# bad - boolean without default
+add_column :users, :active, :boolean
+
+# good - boolean with default
+add_column :users, :active, :boolean, default: true
```
pirj, line 1028: It might default to `true`, but it's still a three-state boolean, i.e. `User.create(active: nil).reload.active # => nil`.

tejasbubane: Right. I missed the `null: false` constraint. Fixed now.

README.adoc
tejasbubane: Added example

README.adoc
tejasbubane: Mentioned `SQL databases`

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
 While enforcing table defaults only in Rails is suggested by many Rails developers, it's an extremely brittle approach that leaves your data vulnerable to many application bugs.
 And you'll have to consider the fact that most non-trivial apps share a database with other applications, so imposing data integrity from the Rails app is impossible.
 
+=== 3-state Boolean [[three-state-boolean]]
+
+With SQL databases, if a boolean column is not given a default value, it ends up having three possible values: `true`, `false` and `NULL`.
+Boolean operators [work in unexpected ways](https://en.wikipedia.org/wiki/Three-valued_logic) with `NULL`.
+
+eg. In SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
```
pirj, line 1034: `eg. In` -> `For example, in` or `E.g. in`.

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
...
+
+To avoid such situations, boolean columns should always have a default value and not null constraint.
```
pirj, line 1036: `not null` -> `NOT NULL`.

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
...
+
+[source,ruby]
+----
+# bad - boolean without default
+add_column :users, :active, :boolean
+
+# good - boolean with default
```
pirj, line 1043: ...and with restricted `NULL`

pirj (approved):
Thank you!

Added a couple of cosmetic notes.

tejasbubane:
Thanks @pirj I have made the suggested changes.

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
...
+E.g. in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
+
+To avoid such situations, boolean columns should always have a default value and `NOT NULL` constraint.
+
+[source,ruby]
+----
+# bad - boolean without default
+add_column :users, :active, :boolean
+
+# good - boolean with default and restricted with null
```
pirj, line 1043: Sorry for nitpicking, but I'm not too confident if "restricted with null" explains what `null: false` does.
I think `with restricted NULL`. A quick search over https://guides.rubyonrails.org/active_record_migrations.html didn't yield any better wording, unfortunately.

tejasbubane: changed

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
 While enforcing table defaults only in Rails is suggested by many Rails developers, it's an extremely brittle approach that leaves your data vulnerable to many application bugs.
 And you'll have to consider the fact that most non-trivial apps share a database with other applications, so imposing data integrity from the Rails app is impossible.
 
+=== 3-state Boolean [[three-state-boolean]]
+
+With SQL databases, if a boolean column is not given a default value, it ends up having three possible values: `true`, `false` and `NULL`.
+Boolean operators [work in unexpected ways](https://en.wikipedia.org/wiki/Three-valued_logic) with `NULL`.
+
+E.g. in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
```
andyw8, line 1034: Minor, but starting a sentence with "E.g." feels awkward. How about "For example"?

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
 While enforcing table defaults only in Rails is suggested by many Rails developers, it's an extremely brittle approach that leaves your data vulnerable to many application bugs.
 And you'll have to consider the fact that most non-trivial apps share a database with other applications, so imposing data integrity from the Rails app is impossible.
 
+=== 3-state Boolean [[three-state-boolean]]
+
+With SQL databases, if a boolean column is not given a default value, it ends up having three possible values: `true`, `false` and `NULL`.
```
andyw8, line 1031: I think "will have" rather than "ends up having" reads better.

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
...
+Boolean operators [work in unexpected ways](https://en.wikipedia.org/wiki/Three-valued_logic) with `NULL`.
+
+E.g. in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
+
+To avoid such situations, boolean columns should always have a default value and `NOT NULL` constraint.
```
andyw8, line 1036: "and **a** `NOT NULL`"

andyw8 (approved):

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
...
+
+[source,ruby]
+----
+# bad - boolean without a default value
+add_column :users, :active, :boolean
+
+# good - boolean with a default value and with restricted NULL
+add_column :users, :active, :boolean, default: true, null: false
```
koic, line 1044: I'd prefer an example code of the rule that defaults to `false` here :-) Of course, In practice, `false` or `true` is chosen depending on the context.

README.adoc
```diff
@@ -1026,6 +1026,24 @@ end
...
+E.g. in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
+
+To avoid such situations, boolean columns should always have a default value and `NOT NULL` constraint.
+
+[source,ruby]
+----
+# bad - boolean without a default value
+add_column :users, :active, :boolean
+
+# good - boolean with a default value and with restricted NULL
```
koic, line 1043:
```suggestion
# good - boolean with a default value (`false` or `true`) and with restricted NULL
```

README.adoc
koic: Or I think one approach is to exemplify an example with `default: false` and an example with `default: true`.

tejasbubane:
@andyw8 @koic Thanks. I have made the suggested changes.

koic:
Thanks!

PR #276, opened 2020-11-27
Recommend ActiveModel native modules over ActiveAttr gem

Drenmi:
For models which don't need to be backed by a database, the built-in ActiveModel modules work perfectly fine, so we don't need to recommend a 3rd party gem.

E.g. validations are already mixed into `ActiveModel::Model`, and starting with modern versions of rails the attributes API is supported through `ActiveModel::Attributes`. For earlier versions, a simple `#attr_accessor` declaration does the trick.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -340,21 +340,16 @@ Introduce non-Active Record model classes freely.
 
 Name the models with meaningful (but short) names without abbreviations.
 
-=== ActiveAttr Gem [[activeattr-gem]]
+=== Non-ActiveRecord Models [[non-activerecord-models]]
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -342,27 +342,39 @@ Name the models with meaningful (but short) names without abbreviations.
 
 === ActiveAttr Gem [[activeattr-gem]]
 
-If you need model objects that support Active Record behavior (like validation) without the Active Record database functionality use the https://github.com/cgriego/active_attr[ActiveAttr] gem.
+If you need objects that support ActiveRecord-like behavior (like validations) without the database functionality, use `ActiveModel::Model`.
```
bbatsov, line 345: I think you should also update the title here.

Drenmi: You are correct. Fixed.

PR #279, opened 2021-02-20
Add a guideline to load config defaults

### LINKED ISSUE #277: Missing `config.load_defaults`
It took me by surprise that if you don't explicitly specify:
```ruby
config.load_defaults 5.2
```
on a Rails 5.2 project, **no** defaults will be loaded, e.g. `config.active_record.belongs_to_required_by_default` will be `nil` and will be treated as `false`. It is `true` since Rails 5.0.

Not specific to Rails 5.2.

Is it worth it adding to the guide?

pirj:
Not loading the defaults will set you back on important Rails framework defaults, e.g.:
- action_view.default_enforce_utf8
- action_controller.default_protect_from_forgery
- active_record.belongs_to_required_by_default

and is sometimes hard to notice.

https://guides.rubyonrails.org/configuring.html#results-of-config-load-defaults

Fixes #277

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -99,6 +99,16 @@ config.assets.precompile += %w( rails_admin/rails_admin.css rails_admin/rails_ad
 
 Keep configuration that's applicable to all environments in the `config/application.rb` file.
 
+=== Load Rails Config Defaults [[config-defaults]]
+
+When upgrading to a newer Rails version, your application's configuration setting will remain on the previous version. To take advantage of the latest recommended Rails practices, the `config.load_defaults` setting should match your Rails version.
+
+[source,ruby]
+----
+# good
+config.load_defaults 6.1
+----
+
 === Staging Like Prod [[staging-like-prod]]
 
 Create an additional `staging` environment that closely resembles the `production` one.


README.adoc
```diff
@@ -99,6 +99,17 @@ config.assets.precompile += %w( rails_admin/rails_admin.css rails_admin/rails_ad
 
 Keep configuration that's applicable to all environments in the `config/application.rb` file.
 
+=== Load Rails Config Defaults [[config-defaults]]
+
+Load Rails' configuration defaults in the `config/application.rb` corresponding to the used Rails version.
```
andyw8, line 104: I think a little more guidance/explanation can be given here:
```suggestion
When upgrading to a newer Rails version, your application's configuration setting will remain on the previous version. To take advantage of the latest recommended Rails practices, the `config.load_defaults` setting should match your Rails version.
```

README.adoc
```diff
@@ -99,6 +99,17 @@ config.assets.precompile += %w( rails_admin/rails_admin.css rails_admin/rails_ad
...
+
+[source,ruby]
+----
+# good
+config.load_defaults 6.1
+----
+
+# good
```
andyw8, line 112: Acciental?

README.adoc
pirj: Significantly better said, thank you!

PR #280, opened 2021-02-21
Discourage using dedicated staging environment

### LINKED ISSUE #240: Avoid creating additional environments?
> Create an additional `staging` environment that closely resembles the `production` one.

It sounds like this is advocating for adding a new config with `config/environments`. This is counter to the advice in https://12factor.net/config

> In a twelve-factor app, env vars are granular controls, each fully orthogonal to other env vars. They are never grouped together as “environments”, but instead are independently managed for each deploy. This is a model that scales up smoothly as the app naturally expands into more deploys over its lifetime.

From experience, when there are multiple environments (beyond the base 3), they often diverge, which lead to problems.

pirj:
Fixes #240

See:
- https://12factor.net/dev-prod-parity
- https://12factor.net/config

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -111,7 +111,8 @@ config.load_defaults 6.1
 
 === Staging Like Prod [[staging-like-prod]]
 
-Create an additional `staging` environment that closely resembles the `production` one.
+Avoid creating additional environment configurations than the defaults of `development`, `test` and `production`.
+If you need a production-like environment such as staging, use environment variables for configuration options.
 
 === YAML Config [[yaml-config]]
 


README.adoc
```diff
@@ -111,7 +111,8 @@ config.load_defaults 6.1
 
 === Staging Like Prod [[staging-like-prod]]
 
-Create an additional `staging` environment that closely resembles the `production` one.
+Avoid creating an additional `staging` environment, use `production`.
```
andyw8, line 114:
```suggestion
Avoid creating additional environment configurations than the defaults of `development`, `test` and `production`.
```

pirj: Thanks!

README.adoc
```diff
@@ -111,7 +111,8 @@ config.load_defaults 6.1
...
+If certain configuration options for your staging server are different, configure them using environment variables.
```
andyw8, line 115:
```suggestion
If you need a production-like environment such as staging, use environment variables for configuration options.
```

README.adoc
andyw8:
```suggestion
Avoid creating additional environment configurations other than the defaults of `development`, `test` and `production`.
```

andyw8:
@pirj sorry, I meant to add my text as a suggestion rather than an edit.

andyw8:
@pirj Oh, you merged so quickly, I didn't realise!

andyw8 (approved):

pirj:
It doesn't take long to click a button 😄 
Thanks for the editing!

PR #281, opened 2021-02-22
Add Custom Validation Methods naming guideline

### LINKED ISSUE #223: Name custom validation methods with prefix 'validate_'
Unless you include the word 'validate' in your method name for a custom validation method, you pretty much ensure that it has a weird name, or an unexpected side effect.

For example:

```ruby
# Bad
class User < ActiveRecord::Base
  validates :email, presence: true
  validate :email_is_not_gmail_account
  #...
  def email_is_not_gmail_account
     # This method name implies that (maybe?) it returns true/false,
     # not that it mutates the errors attribute of this model.
  end
end

# Good
class User < ActiveRecord::Base
  validates :email, presence: true
  validate :validate_email_is_not_gmail_account
  #...
  def validate_email_is_not_gmail_account
     # This method name tells you exactly what it does
  end
end

```

pirj:
Fixes #223

Better viewed as https://github.com/rubocop-hq/rails-style-guide/blob/custom-validation-method-naming/README.adoc#custom-validation-methods

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -552,6 +552,32 @@ validates_length_of :email, maximum: 100
 validates :email, presence: true, length: { maximum: 100 }
 ----
 
+=== Custom Validation Methods
+
+When naming custom validation methods, adhere to the simple rules:
+
+ - `validate :method_name` reads like a natural statement
+ - the method name explains what it checks
+ - the method is recognizable as a validation method by its name, not a predicate method
+
+[source,ruby]
+----
+# good
+validate :expiration_date_cannot_be_in_the_past
+validate :discount_cannot_be_greater_than_total_value
+validate :ensure_same_topic
+
+# also good - explicit prefix
+validate :validate_birthday_in_past
+validate :validate_sufficient_quantity
+validate :must_have_owner_with_no_other_items
+validate :must_have_shipping_units
+
+# bad
+validate :birthday_in_past
+validate :owner_has_no_other_items
+----
+
 === Single-attribute Validations [[single-attribute-validations]]
 
 To make validations easy to read, don't list multiple attributes per validation.
--- a/README.adoc
+++ b/README.adoc
@@ -565,7 +565,7 @@ When naming custom validation methods, adhere to the simple rules:
 # good
 validate :expiration_date_cannot_be_in_the_past
 validate :discount_cannot_be_greater_than_total_value
-validate :ensure_same_topic
+validate :ensure_same_topic_is_chosen
 
 # also good - explicit prefix
 validate :validate_birthday_in_past


README.adoc
```diff
@@ -552,6 +552,33 @@ validates_length_of :email, maximum: 100
...
+ - the method name explains what it checks
+ - the method is recognizable as a validation method by its name, not a predicate method
+
+[source,ruby]
+----
+# good
+validate :expiration_date_cannot_be_in_the_past
+validate :discount_cannot_be_greater_than_total_value
+validate :ensure_same_topic
+validate :referrer_valid
```
yjukaku, line 569: This might not be a good example - it sounds like a predicate method. The others make sense to me though. Maybe we remove this one line?

yjukaku (commented):
Thanks so much for putting up this PR! I think it pretty much sums up the conversation we were having in #223.

README.adoc
pirj: Thanks for the heads-up! I was under the impression that I've removed this one in the last proofreading sweep.

README.adoc
```diff
@@ -552,6 +552,32 @@ validates_length_of :email, maximum: 100
...
+ - `validate :method_name` reads like a natural statement
+ - the method name explains what it checks
+ - the method is recognizable as a validation method by its name, not a predicate method
+
+[source,ruby]
+----
+# good
+validate :expiration_date_cannot_be_in_the_past
+validate :discount_cannot_be_greater_than_total_value
+validate :ensure_same_topic
```
andyw8, line 568: This last one doesn't read as naturally to me... maybe `validate :same_topic_is_chosen`?

README.adoc
```diff
@@ -552,6 +552,32 @@ validates_length_of :email, maximum: 100
...
+
+# also good - explicit prefix
+validate :validate_birthday_in_past
```
andyw8, line 571: Repeating `validate_` in the name seems a bit awkward...

README.adoc
```diff
@@ -552,6 +552,32 @@ validates_length_of :email, maximum: 100
...
+validate :validate_sufficient_quantity
+validate :must_have_owner_with_no_other_items
+validate :must_have_shipping_units
+
+# bad
+validate :birthday_in_past
```
andyw8, line 577: Neither of these seem that bad... can we find some worse examples? 😁

README.adoc
pirj: Indeed. However, it seems it's a [widespread practice](https://github.com/rubocop-hq/rails-style-guide/issues/223#issuecomment-358597186).
[Here's an explanation](https://github.com/rubocop-hq/rails-style-guide/issues/223#issuecomment-522912340):

> methods are what makes up the API that's going to be reused. If I'm scanning through the method list and see something like name_must_be_snake_case, I don't immediately know what that method does. It could be used for validation but it could also return a boolean that indicates if name must be snake_case or not. If that method were called validate_name_is_snake_case, I'd know at once that it's not something I'd want to reuse for anything other than validation.

I don't feel like to be using this variation myself, but we don't insist on using it and provide the freedom to use other naming options.

README.adoc
pirj: Maybe `ensure_same_topic_is_chosen`, WDYT? Without `ensure` it really looks like a predicate.

README.adoc
pirj: The worst custom validation method names I could find:
```ruby
validate :final_grader_ok?
validate :valid_role?
validate :positive_points_possible?
validate :published_article
validate :users_for_shared_access
validate :close_after_open
validate :orphaned_annotation
validate :not_from_fork
validate :one_issue_tracker
validate :first_default_scope
validate :hosts_format
validate :permissions
```

Some of them read very well in conjunction with `validate'. But method names themselves don't give a clue that it's a custom validation method.

Which of them do you dislike most of all?

yjukaku: The issue that I was raising in #223  was not that the validation methods read poorly when being called as part of the `validate :xyz` line, but rather that reading through the class's code doesn't make it immediately obvious if the method is a validation method or a predicate method (or something else). That's why I personally have adopted the `validate :validate_xyz` pattern, even though admittedly it reads somewhat poorly.

README.adoc
andyw8: My 'test' was to read it like a set of business rules and prefix each with "we":

"We validate expiration date cannot be in the past"
"We validate discount code cannot be total value"
"We validate ensure same topic" <- doesn't seem to flow

pirj: We have several good sections with different naming patterns. The two in this section read well with "we", as they both are taken from the same source. The third one is from a different source.

Do you feel that it's necessary to have more good sections or more diversity in the primary good section?

andyw8 (approved):

pirj:
@koic Does this guideline look good to you?

savariya28 (approved):

pirj:
I'll merge this. We can improve it basing on the feedback that is more likely to happen when it's published.

PR #286, opened 2021-05-09
Use `email_address_with_name`

andyw8:
https://boringrails.com/tips/rails-mailers-email-address-with-name

Hat tip to @swanson

PR #288, opened 2021-08-08
Change params order for email_address_with_name method

georgiybykov:
Hi!

According to the [docs](https://apidock.com/rails/ActionMailer/Base/email_address_with_name/class) we have to pass the email by first parameter to the method and the name as a second one.
Otherwise we will catch an `Mail::Field::IncompleteParseError`.

Best regards,
Georgiy


andyw8:
Yes, sorry that was my mistake in #286.

andyw8 (approved):

pirj:
Thank you!

PR #294, opened 2021-10-10
Add examples for enforcing FKs

### LINKED ISSUE #287: Use `add_reference`/`t.references` in migrations with `foreign_key: true`
According to https://api.rubyonrails.org/v6.1/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference, it defaults to `false`.

When `dependent:` on `belongs_to` is not properly set, it may leave orphaned records behind on association deletion.

It also prevents from early detection of code responsible for referral integrity violations.

pirj:
fixes #287

Rendered: https://github.com/rubocop/rails-style-guide/blob/add-example-for-fks/README.adoc#foreign-key-constraints

Rails docs: https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_reference

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1086,6 +1086,23 @@ add_column :users, :admin, :boolean, default: false, null: false
 
 Enforce foreign-key constraints. As of Rails 4.2, Active Record supports foreign key constraints natively.
 
+[source,ruby]
+----
+# bad - does not add foreign keys
+create_table :comment do |t|
+  t.references :article
+  t.belongs_to :user
+  t.integer :category_id
+end
+
+# good
+create_table :comment do |t|
+  t.references :article, foreign_key: true
+  t.belongs_to :user, foreign_key: true
+  t.references :category, foreign_key: { to_table: :comment_categories }
+end
+----
+
 === Change vs Up/Down [[change-vs-up-down]]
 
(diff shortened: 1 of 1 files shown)


andyw8:
To help beginners, I wonder if this should include another integer field which _shouldn't_ have a foreign key (e.g. `position`). That would help illustrate that not every fields needs to be declared as a foreign key.

pirj:
@andyw8 Something like this?
```ruby
# bad - does not add foreign keys
create_table :comment do |t|
  t.string :email # does not need a foreign key constraint
  t.references :article
  t.belongs_to :user
  t.integer :category_id
end
# good
create_table :comment do |t|
  t.string :email # does not need a foreign key constraint
  t.references :article, foreign_key: true
  t.belongs_to :user, foreign_key: true
  t.references :category, foreign_key: { to_table: :comment_categories }
end
```

I would be troubled to understand, how is email related to FKs guideline. On the other hand, some non-integer `status` column may deserve to be a FK if it's a natural PK in `statuses` table. And it needs `on_delete: :nullify`/`on_update: :cascade`. I just don't want to dive that deep in this guideline.

PR #296, opened 2021-11-20
Add section for `where.not` with multiple attributes

tejasbubane:
Issue:https://github.com/rubocop/rubocop-rails/issues/565

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1086,6 +1086,21 @@ Customer.where("purchases_count > :min AND purchases_count <= :max", min: 0, max
 
 NOTE: Rails 6.0 or later is required for endless range Ruby 2.6 syntax, and Rails 6.0.3 for beginless range Ruby 2.7 syntax.
 
+=== `where.not` with multiple attributes
+
+Avoid passing multiple attributes to `where.not`. Rails logic in this case has changed in Rails 6.1 and
+will now yield results matching either of those conditions,
+e.g. `where.not(status: 'active', plan: 'basic')` would return records with active status when the plan is business.
+
+[source, ruby]
+----
+# bad
+User.where.not(status: 'active', plan: 'basic')
+
+# good
+User.where.not('status = ? AND plan = ?', 'active', 'basic')
+----
+
 == Migrations
 
 === Schema Version [[schema-version]]


README.adoc
```diff
@@ -1029,6 +1029,20 @@ User.all.size
 User.all.length
 ----
 
+=== `where.not` with multiple attributes
+
+Until Rails 6.1, when negating condition with multiple attributes using `where.not`, it works like NOR (NOT(A) AND NOT(B)). This does not follow boolean algebra law `NOT (A AND B) = (NOT(A) OR NOT(B))` and could lead to confusing results.
```
pirj, line 1034: Is the example with `NOR` correct here? It operates on two values.

tejasbubane: I meant `NOT(A) AND NOT(B)` is `NOR`. Anyways I think the sentence is better off without mentioning `NOR` and `NAND` to avoid confusion. Removed it.

README.adoc
```diff
@@ -1029,6 +1029,20 @@ User.all.size
 User.all.length
 ----
 
+=== `where.not` with multiple attributes
+
+Until Rails 6.1, when negating condition with multiple attributes using `where.not`, it works like `(NOT(A) AND NOT(B))`. This does not follow boolean algebra law `NOT (A AND B) = (NOT(A) OR NOT(B))` and could lead to confusing results.
```
koic, line 1034: Until Rails 6.0?
https://github.com/rails/rails/pull/36029

tejasbubane: `6.0` added a deprecation warning but it still worked like NOR. The switch happened in `6.1` as the commit message describes: https://github.com/rails/rails/pull/36029/commits/12a9664ff60f0e2712fd1f79f8dbec06e2f004a2

pirj:
I find the [original reasoning](https://github.com/rails/rails/issues/31209) confusing:
> Expected behavior
> NOT(A AND B) = NOT(A) OR NOT(B)

To me, both `where.not(a: 1, b: 2)` and `where.not(a: 1).where.not(b: 2)` should be `NOT(A) AND NOT(B)`.

I agree with the bad example:
```ruby
# bad
User.where.not(status: "active", plan: "basic")
```

But I don't agree with the message involving boolean algebra reasoning.
I'd say:
> Avoid passing multiple attributes to a `where.not`, as Rails logic in this case has changed in Rails 6.1 and will now yield results matching either of those conditions, e.g. `where.not(status: 'active', plan: 'basic')` would return records with `active` `status` when the `plan` is `business`.

@andyw8 WDYT?

andyw8:
Personally I would avoid `not` except for trivial cases with a single attribute, but I also wouldn't use a chained `not` as it can still be highly confusing. Instead I would write it as SQL to be explicit about the intent.

pirj:
I side with @andyw8 and would recommend not using `where.not` with multiple attributes or chained.

The few real-world usages of `where.not` with multiple attributes I could find in `real-world-rspec` repos:
```
open-source-billing/app/helpers/application_helper.rb
415:    PublicActivity::Activity.where.not(owner_id: current_user.id, key: 'client.update').where(is_read: false).count
423:    PublicActivity::Activity.where.not(owner_id: current_user.id, key: 'client.update').order('created_at desc').page(1).per(10) if current_user.present?

open-source-billing/app/controllers/notifications_controller.rb
3:    @activities = PublicActivity::Activity.where.not(owner_id: current_user.id, key: 'client.update').order('created_at desc').page(params[:page].to_i + 1).per(10) if current_use
r.present?

gitlabhq/app/models/concerns/milestoneable.rb
22:    scope :without_particular_release, -> (tag, project_id) { joins_milestone_releases.where.not(milestones: { releases: { tag: tag, project_id: project_id } }) }
```

I have no confidence regarding chained `where.not(...).where.not(...)` calls, but I imagine there should be even less of them in practice.

WDYT of discouraging using it? @tejasbubane Would you adjust the PR?

pirj:
Ping @tejasbubane

pirj:
Ping

pirj:
Ping @tejasbubane

tejasbubane:
@pirj Sorry for the long delay here, I've updated as per suggestions.

README.adoc
```diff
@@ -1086,6 +1086,21 @@ Customer.where("purchases_count > :min AND purchases_count <= :max", min: 0, max
...
+will now yield results matching either of those conditions,
+e.g. `where.not(status: 'active', plan: 'basic')` would return records with active status when the plan is business.
+
+[source, ruby]
+----
+# bad
+User.where.not(status: "active", plan: "basic")
+
+# good
+User.where.not("status = ? AND plan = ?", "active", "basic")
```
pirj, line 1101: `NOT (... AND ...)` is a direct replacement for `where.not` with multiple conditions, i.e. it would return records with active statu on the business plan.

The cop [suggests](https://github.com/rubocop/rubocop-rails/commit/7d03a57db63e3a89e723082705561f75304a4d21#diff-9074cb50add7c36de7890a2002f6f1d11b5bb582594ebc75b714524b8ebd33e7R23-R24) multiple chained statements and a `NOT (... OR ...)` to match the older Rails behaviour.
But suggesting this would need a clarification.

pirj (approved):
Thank you!

README.adoc
```diff
@@ -1086,6 +1086,21 @@ Customer.where("purchases_count > :min AND purchases_count <= :max", min: 0, max
...
+=== `where.not` with multiple attributes
+
+Avoid passing multiple attributes to `where.not`. Rails logic in this case has changed in Rails 6.1 and
+will now yield results matching either of those conditions,
+e.g. `where.not(status: 'active', plan: 'basic')` would return records with active status when the plan is business.
+
+[source, ruby]
+----
+# bad
+User.where.not(status: "active", plan: "basic")
```
koic, line 1098: This example could be consistent with single quotes.

```suggestion
User.where.not(status: 'active', plan: 'basic')
```

The following good example is the same.

koic:
Thanks!

PR #298, opened 2021-12-23
Fix parentheses for Squished Heredocs example

xhs345:


pirj:
It is fine as is, surprisingly enough [heredocs work that way](https://github.com/rubocop/rubocop-rspec/blob/573f22bbbd54e5332231b9823b87843550ae21a4/spec/rubocop/cop/rspec/context_method_spec.rb#L5).

PR #299, opened 2021-12-25
Relax dot-separated locale keys guideline

pirj:
Along the way, clarify [Lazy lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup) guideline a bit:

"the texts" is quite vague, as it stands for:
> strings and other locale specific bits (such as date or currency formats)

I believe "entries" is a better term.

Lazy lookup is also available in controllers, see https://github.com/rails/rails/blob/fee61e3abc878e4d4930a0be0accf8e4569009a8/actionpack/lib/abstract_controller/translation.rb#L11
PS Except `ActionController::Api`. Only `ActionController::Base`.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1326,7 +1326,7 @@ Use the short form of the I18n methods: `I18n.t` instead of `I18n.translate` and
 
 === Lazy Lookup [[lazy-lookup]]
 
-Use "lazy" lookup for the texts used in views. Let's say we have the following structure:
+Use "lazy" lookup for locale entries from views and controllers. Let's say we have the following structure:
 
 ----
 en:
@@ -1339,6 +1339,10 @@ The value for `users.show.title` can be looked up in the template `app/views/use
 
 [source,ruby]
 ----
+# bad
+= t 'users.show.title'
+
+# good
 = t '.title'
 ----
 
--- a/README.adoc
+++ b/README.adoc
@@ -1348,8 +1348,8 @@ The value for `users.show.title` can be looked up in the template `app/views/use
 
 === Dot-separated Keys [[dot-separated-keys]]
 
-Use the dot-separated keys in the controllers and models instead of specifying the `:scope` option.
-The dot-separated call is easier to read and trace the hierarchy.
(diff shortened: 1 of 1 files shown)


pirj:
@rubocop/style-guide-editors Ping

README.adoc
```diff
@@ -1330,7 +1330,7 @@ Use the short form of the I18n methods: `I18n.t` instead of `I18n.translate` and
 
 === Lazy Lookup [[lazy-lookup]]
 
-Use "lazy" lookup for the texts used in views. Let's say we have the following structure:
+Use "lazy" lookup for locale entries from views and controllers. Let's say we have the following structure:
```
andyw8, line 1333: Helpers too?

README.adoc
```diff
@@ -1343,21 +1343,27 @@ The value for `users.show.title` can be looked up in the template `app/views/use
...
 
 [source,ruby]
 ----
 # bad
 I18n.t :record_invalid, scope: [:activerecord, :errors, :messages]
+I18n.t :title, scope: :invitation
 
 # good
 I18n.t 'activerecord.errors.messages.record_invalid'
+I18n.t :record_invalid, scope: 'activerecord.errors.messages'
```
andyw8, line 1366: Maybe each example should correspond to a `bad` entry?

README.adoc
pirj: I believe this may lead to confusion.
The term "helpers" can be used for ["form helpers"](https://guides.rubyonrails.org/form_helpers.html), ["view helpers"](https://guides.rubyonrails.org/action_view_helpers.html), [ad-hoc helpers](https://github.com/spree/spree/blob/main/core/app/helpers/spree/locale_helper.rb#L17), and even [email helpers](https://github.com/discourse/discourse/blob/bebb4fd11181f3deb277b2e4cb0e7148b61fb71d/app/helpers/email_helper.rb#L20). I have no certainty that it is possible to use lazy lookup in all those cases.

README.adoc
pirj: Fixed, thanks 👍

andyw8 (approved):

PR #300, opened 2022-01-05
Fix code highlighting issues

cuchi:
This little change fixes the code highlighting in two places, which only seems to happen on the generated doc (when using asciidoctor + rouge):

2022-01-05-101104_1015x399_scrot


koic:
Nice 👍

PR #301, opened 2022-01-05
Prefer appending to setting `ignored_columns`

fsateler:
Setting may overwrite previous assignments and that is almost always a mistake.

See also https://github.com/rubocop/rubocop-rails/pull/514

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -404,6 +404,22 @@ class Transaction < ApplicationRecord
 end
 ----
 
+=== Always append to `ignored_columns` [[append-ignored-columns]]
+
+Avoid setting `ignored_columns`. It may overwrite previous assignments and that is almost always a mistake. Prefer appending to the list instead.
+
+[source,ruby]
+----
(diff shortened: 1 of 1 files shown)


pirj (approved):
Looks both reasonable and safe.
Thank you!

Just as a matter of explanation. Overwriting the setting set in a base class (think STI) or an included module leads to certain columns are not actually ignored. This can result in a false confidence that certain DB columns are not in the cached schema and are not requested along from the DB, causing the app to fail with a downtime.

pirj:
@rubocop/style-guide-editors Ping

pirj:
To be fair, the only project where this technique is used in the wild is [GitLab](https://github.com/gitlabhq/gitlabhq/blob/17ef30f3df6d3939e41e69efc7cfa3deaa08605d/app/models/concerns/ignorable_columns.rb#L24)

fsateler:
Thanks!

PR #305, opened 2022-02-05
Use consistent language for `prefer` vs `favor`

andyw8:
There are 10 uses of `prefer...over` but only 4 uses of `favor...over`.

(Using `prefer` also avoids having to deal with the different spellings `favor`).

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -871,7 +871,7 @@ Client.where(
 
 === `find` [[find]]
 
-Favor the use of `find` over `where.take!`, `find_by!`, and `find_by_id!` when you need to retrieve a single record by primary key id and raise `ActiveRecord::RecordNotFound` when the record is not found.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -871,7 +871,7 @@ Client.where(
 
 === `find` [[find]]
 
-Favor the use of `find` over `where.take!`, `find_by!`, and `find_by_id!` when you need to retrieve a single record by primary key id and raise `ActiveRecord::RecordNotFound` when the record is not found.
+Prefer the use of `find` over `where.take!`, `find_by!`, and `find_by_id!` when you need to retrieve a single record by primary key id and raise `ActiveRecord::RecordNotFound` when the record is not found.
```
pirj, line 874: Just thinking aloud.
Does it make sense to use passages like "Prefer to use", wouldn't just "Use" do?

andyw8: Possibly? That should be a wider discussion, as similar language is used in the Ruby Style Guide.

andyw8: Actually, `the use of` does seem unnecessary, I will trim that.

pirj (approved):
Looks good, thanks!

PR #306, opened 2022-02-05
Add original author credit

andyw8:
The introduction uses "I", but as the guide is no longer under Bozhidar's account, it's not clear who this refers to.


pirj (approved):
The "I didn't come up" made me think of who the original author was for a second just today.
Thanks!

PR #308, opened 2022-02-06
Remove outdated platform advice

andyw8:
I expect this was added before Bundler had proper support for multiple platforms.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1705,32 +1705,6 @@ Put gems used only for development or testing in the appropriate group in the Ge
 Use only established gems in your projects.
 If you're contemplating on including some little-known gem you should do a careful review of its source code first.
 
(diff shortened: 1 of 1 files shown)


andyw8:
(was originally added by @sirupsen in https://github.com/rubocop/rails-style-guide/commit/eb3520eb451e56359181df9fef83ed72a19490f5 in 2011).

pirj (approved):
I see it's completely fine to drop this.
Do we need to mention the modern way to do so? E.g.
```ruby
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```
I have no certainty that the above is correct, I took it from a project that is never used on any of those platforms.

andyw8:
The above matches what Rails generates on a new project.

I don't think there's a need to give any specific guidance here.

PR #309, opened 2022-02-06
Remove stale translations

andyw8:
I think we need to have some kind of cut-off for translations that are no longer updated. I'm proposing we remove the ones that haven't been updated for over 5 years.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -59,14 +59,8 @@ gem install rouge
 
 Translations of the guide are available in the following languages:
 
-* https://github.com/JuanitoFatas/rails-style-guide/blob/master/README-zhCN.md[Chinese Simplified]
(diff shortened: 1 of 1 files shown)


andyw8:
cc @JuanitoFatas @tolgaavci @pureugong @CQBinh @abraaomiranda

pirj (approved):
This problem also affects the Ruby Style Guide.
I see more harm than benefits to have stale localized guides.
The only one that is more or less up to date is the [Russian Ruby SG](https://github.com/arbox/ruby-style-guide/blob/master/README-ruRU.md), and still it's one year behind the main guide.
We lack resources to sync every single localization.
And if we did, from my perspective, those guides should all be in the same repo, and localizations should be kept on different branches.

👍 for removal of references to stale guides, and I would go one step further and removed them all.

Issue #284, opened 2021-04-09
Clarify the reasoning for Define Model Class Migrations


pirj:
1. https://rails.rubystyle.guide/#define-model-class-migrations claims that

> or cause serious data corruption

Can you think of a use case when an old migration could cause a serious data corruption?
Yes, it will most probably break `db:migrate` on a clean database.

2. There's yet another reason not to use real models in migrations.
If you refer to a real model in an older migration (e.g. for `update_all`), Rails will cache column information.
And subsequent migrations that would alter that table's schema won't affect the column cache.
This will lead to breakages due to an outdated (fixed for the moment when the model was first referred to in migrations) table schema. Specifically, if you add a column in a later migration and `update_all` that new column, this will result in an error.
Also, if you run `rake db:migrate db:seed` in one command, this will also affect your seeds, and it will also fail with:
```
ActiveModel::UnknownAttributeError: unknown attribute 'some_attr' for ModelName
```

andyw8:
For reference, this was originally added in https://github.com/rubocop/rails-style-guide/pull/205

pirj:
There are two more things that recently hit me:

3. We suggest defining the class on the top level:
```ruby
class MigrationProduct < ActiveRecord::Base
  self.table_name = :products
end

class ModifyDefaultStatusForProducts < ActiveRecord::Migration
```

while the `MigrationProduct` can be defined by multiple migrations.
If it happens that `self.table_name = ` is different, the last one wins.
I don't have a good example, maybe something like:
```ruby
class MigrationUser < ActiveRecord::Base
  self.table_name = 'accounts' # not 'users'
end
```

The suggestion would be to define migration model classes inside migration classes:
```ruby
class ModifyDefaultStatusForProducts < ActiveRecord::Migration
  class MigrationProduct < ActiveRecord::Base
    self.table_name = :products
  end
  ...
```

pirj:
4. Migration models are loaded as needed in development, while loaded all at once in production with `eager_load`. (this needs to be confirmed with the Rails code, if the migrator `load`s them or it loads the whole `db/migrate` and instantiates migration classes as needed.

Still, the issue was as following:
```ruby
# migration 1
class AddCategoryToFoos < ActiveRecord::Migration[6.0]
  def change
    add_column :foos, :category, :string
  end
end

# migration 2
class BackfillCategoryToFoos < ActiveRecord::Migration[6.0]
  class MigrationFoo < ActiveRecord::Base
[...]

PR #310, opened 2022-02-06
Replace references to `ActiveRecord::Base` with `ApplicationRecord`

andyw8:
This changed in Rails 5:

https://www.bigbinary.com/blog/application-record-in-rails-5

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -179,11 +179,11 @@ Use nested routes to express better the relationship between Active Record model
 
 [source,ruby]
 ----
-class Post < ActiveRecord::Base
+class Post < ApplicationRecord
   has_many :comments
 end
 
-class Comment < ActiveRecord::Base
+class Comment < ApplicationRecord
   belongs_to :post
 end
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1162,7 +1162,7 @@ end
 # same table you had during the creation of the migration.
 # In future if you override the `Product` class and change the `table_name`,
 # it won't break the migration or cause serious data corruption.
-class MigrationProduct < ActiveRecord::Base
```
pirj, line 1165: I'm on the fence with changing it here, and the reason is https://github.com/rubocop/rails-style-guide/issues/284.
Wondering what enhancements could go into `ApplicationRecord` that would be useful in migrations.
And since those enhancements are a subject to change over time, running historical migrations can be affected by the same issue as with renaming/removing models.
I'd keep this guideline intact, maybe with a clarification.

pirj (approved):
Thanks!
Looks good with a single exception in "Define Model Class Migrations" guideline.

README.adoc
andyw8: I've put this one back to `ActiveRecord::Base`.

PR #317, opened 2022-05-28
Add "Prefer using squiggly heredoc over `strip_heredoc`" rule

koic:
Follow up https://github.com/rubocop/rubocop-rails/pull/706.

This PR adds "Prefer using squiggly heredoc over `strip_heredoc`" rule.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1575,6 +1575,28 @@ hash.exclude?(:key)
 string.exclude?('substring')
 ----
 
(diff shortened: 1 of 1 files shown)


pirj (approved):
Thanks!

Just in case, related [Ruby Style Guide "Squiggly Heredocs" guideline](https://rubystyle.guide/#squiggly-heredocs).
I agree that it makes sense to duplicate it in the Rails Style Guide, as `strip_heredoc` is a Railsism.

PR #318, opened 2022-06-14
Add "Prefer `freeze_time` over `travel_to` with an argument of the current time" rule

ydah:
Since there are several ways to indicate the current time, we believe that `freeze_time`, which can be expressed as simply as possible, should be preferred.

```ruby
# bad
travel_to(Time.now)
travel_to(DateTime.now)
travel_to(Time.current)
travel_to(Time.zone.now)
travel_to(Time.now.in_time_zone)
travel_to(Time.current.to_time)

# good
freeze_time
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1791,6 +1791,24 @@ class MyControllerTest < ActionDispatch::IntegrationTest
 end
 ----
 
+=== `freeze_time` [[freeze-time]]
+
+Prefer https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time[ActiveSupport::Testing::TimeHelpers#freeze_time] over https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to[ActiveSupport::Testing::TimeHelpers#travel_to] with an argument of the current time.
+
+[source,ruby]
+----
+# bad
+travel_to(Time.now)
+travel_to(DateTime.now)
+travel_to(Time.current)
+travel_to(Time.zone.now)
+travel_to(Time.now.in_time_zone)
+travel_to(Time.current.to_time)
+
+# good
+freeze_time
+----
+
 == Managing Processes
 
 === Foreman [[foreman]]


pirj:
It is worth mentioning that `freeze_time` just delegates to `travel_to` with a default `Time.now` (https://github.com/rails/rails/blob/aa3a6c9c1b70e49bef5a2713aae96179debdff3e/activesupport/lib/active_support/testing/time_helpers.rb#L237):
```ruby
      def freeze_time(&block)
        travel_to Time.now, &block
      end
```

README.adoc
```diff
@@ -1791,6 +1791,26 @@ class MyControllerTest < ActionDispatch::IntegrationTest
 end
 ----
 
+== Additional Testing Resources [[additional-testing-resources]]
```
andyw8, line 1794: I think this could just be part of the previous 'Testing' section?

README.adoc
```diff
@@ -1791,6 +1791,26 @@ class MyControllerTest < ActionDispatch::IntegrationTest
...
+
+=== `freeze_time` [[freeze-time]]
+
+Prefer https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time[ActiveSupport::Testing::TimeHelpers#freeze_time] over https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to[ActiveSupport::Testing::TimeHelpers#travel_to] with argument current time.
```
andyw8, line 1798:
```suggestion
Prefer https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-freeze_time[ActiveSupport::Testing::TimeHelpers#freeze_time] over https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to[ActiveSupport::Testing::TimeHelpers#travel_to] with an argument of the current time.
```

README.adoc
andyw8: Oh, I see. Yeah, I think this is better, there's not really any convention of following the Rails Guides structure.

andyw8 (approved):

PR #319, opened 2022-06-25
Add new "Prefer `all_(day|week|month|quarter|year)` over range of date/time" rule

koic:
Follow up https://github.com/rubocop/rubocop-rails/pull/730.

This PR adds new "Prefer `all_(day|week|month|quarter|year)` over range of date/time" rule.

```ruby
# bad
date.beginning_of_day..date.end_of_day
date.beginning_of_week..date.end_of_week
date.beginning_of_month..date.end_of_month
date.beginning_of_quarter..date.end_of_quarter
date.beginning_of_year..date.end_of_year

# good
date.all_day
date.all_week
date.all_month
date.all_quarter
date.all_year
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1723,6 +1723,28 @@ Time.zone.now # => Fri, 12 Mar 2014 22:04:47 EET +02:00
 Time.current # Same thing but shorter.
 ----
 
+=== Prefer `all_(day|week|month|quarter|year)` over range of date/time [[date-time-range]]
+
+Prefer `all_(day|week|month|quarter|year)` over `beginning_of_(day|week|month|quarter|year)..end_of_(day|week|month|quarter|year)`
+to get the range of date/time.
+
+[source,ruby]
+----
+# bad
+date.beginning_of_day..date.end_of_day
+date.beginning_of_week..date.end_of_week
+date.beginning_of_month..date.end_of_month
+date.beginning_of_quarter..date.end_of_quarter
+date.beginning_of_year..date.end_of_year
+
+# good
+date.all_day
+date.all_week
+date.all_month
+date.all_quarter
+date.all_year
+----
+
 == Duration
 
 === Duration Application


pirj:
I ran `rg --no-ignore --hidden "\.all_(day|week|year|quarter|month){1}"` over `real-world-rspec` and found quite a few usages:
```
canvas-lms/lib/dates_overridable.rb
226:      hash[:all_day] = assignment.all_day
227:      hash[:all_day_date] = assignment.all_day_date # ?????

# a dozen more usages in the same repo, but there's `def all_day`, so it's not the Rails one


loomio/vue/src/components/poll/meeting/form.vue
19:          {text: @$t('common.all_day'), value: null}

mastodon/app/helpers/admin/announcements_helper.rb
5:    if announcement.all_day?

mastodon/app/models/announcement.rb
89:    self.all_day = false if starts_at.blank? || ends_at.blank?


# Only two real usages:

gitlabhq/spec/models/user_custom_attribute_spec.rb
31:      subject { UserCustomAttribute.by_updated_at(Date.today.all_day) }

gitlabhq/app/models/user_custom_attribute.rb
36:      by_key('blocked_at').by_updated_at(Date.yesterday.all_day)
```

I love the API, really. But should we recommend something that is not widespread? Maybe start with the cop and see if this good practice spreads?

koic:
These are APIs that are fully compatible with bad cases and are a smart solution to avoid redundant writing.
So it can start with cop, but I think it's useful to mention it in the style guide as well. While it may be less known than I think, but it's useful to know this API.

Especially seen in the totaling up of the data of Rails applications, which may not be very common in OSS :-)

README.adoc
```diff
@@ -1723,6 +1723,28 @@ Time.zone.now # => Fri, 12 Mar 2014 22:04:47 EET +02:00
 Time.current # Same thing but shorter.
 ----
 
+=== Prefer `all_(day|week|month|quarter|year)` over range of date/time [[date-time-range]]
+
+Prefer `all_(day|week|month|quarter|year)` over `beginning_of_(day|week|month|quarter|year)..end_of_(day|week|month|quarter|year)`
+to get the range of date/time.
```
pirj, line 1729: "the range of date/time" -> "period"?
I understand that the current wording is technical, and the proposed is semantic.
Leaving it up to you what to go with.

pirj (approved):
Agree with your points.
Thanks!

README.adoc
koic: I chose `date/time` from the following documentation, but I think `period` will look like.
https://github.com/rails/rails/blob/v7.0.3/activesupport/lib/active_support/core_ext/date_and_time/calculations.rb#L299-L323

OTOH, this wording `date/time` is matched with the name of `Rails/DateTimeRange` cop.
https://github.com/rubocop/rubocop-rails/pull/730

It's always difficult to name 😅

PR #320, opened 2022-06-28
Add "Finding missing relationship records" rule

ydah:
If you're using Rails 6.1 or higher, prefer `where.missing` over `left_joins` and `where` to find missing relationship records.
It can be expressed in a more simplified.

```ruby
# bad
Post.left_joins(:author).where(authors: { id: nil })

# good
Post.where.missing(:author)
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -938,6 +938,19 @@ User.where("id != ?", id)
 User.where.not(id: id)
 ----
 
+=== Finding missing relationship records [[finding-missing-relationship-records]]
+
+If you're using Rails 6.1 or higher, use https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-missing[where.missing] to find missing relationship records.
+
+[source,ruby]
+----
+# bad
+Post.left_joins(:author).where(authors: { id: nil })
+
+# good
+Post.where.missing(:author)
+----
+
 === Order by `id` [[order-by-id]]
 
 Don't use the `id` column for ordering.


koic:
It would be better to mention requiring Rails 6.1 or higher.
https://github.com/rails/rails/commit/9bfe89e68ee61d9c8ed71ac98c783a7e47bdb608

ydah:
Indeed it is, I added the supplement that it is for Rails 6.1 or higher.

README.adoc
```diff
@@ -938,6 +938,19 @@ User.where("id != ?", id)
 User.where.not(id: id)
 ----
 
+=== Finding orphan records [[finding-orphan-records]]
+
+If you're using Rails 6.1 or higher, use https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-missing[where.missing] to find orphan records.
```
pirj, line 943: I would change the phrasing slightly.
Those are not always orphan records, implying that there's a deficiency in the FKs in the DB schema.

There could be legitimate cases for a record not to have an associated record, e.g. order->delivery (no delivery scheduled because the order is not paid yet), node->parent (root node), post->comments (no comments yet).

pirj (approved):
Thanks!

README.adoc
ydah: Indeed it is. How about "Finding records with nothing associated records" for a different expression?

andyw8: How about mirroring the terminology from the [docs](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-missing), e.g.

> to identify missing relations.

(you could argue that 'missing' also suggests a deficiency, but I think it's better than 'orphaned').

ydah: I think it looks good.
I changed it, how do you like it?

README.adoc
```diff
@@ -938,6 +938,19 @@ User.where("id != ?", id)
 User.where.not(id: id)
 ----
 
+=== Finding identify missing relationship records [[finding-identify-missing-relationship-records]]
+
+If you're using Rails 6.1 or higher, use https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods/WhereChain.html#method-i-missing[where.missing] to find identify missing relationship records.
```
pirj, line 943: "find identify" -> just "find"?

pirj (approved):
Thank you!

pirj (approved):

andyw8 (approved):

koic:
Thanks!

PR #322, opened 2022-07-02
Correct endless range example and provide some clarity

### LINKED ISSUE #321: Where with Ranges: "good" vs "bad" are not equivalent
Using an infinite range actually results in an inclusive query most of the time. Mirroring the documentation (using a "Client" model since my example app doesn't have "User")

```ruby
Client.where("created_at > ?", 7.days.ago).to_sql
# "SELECT \"clients\".* FROM \"clients\" WHERE (created_at > '2022-06-24 11:00:58.795520')"
Client.where(created_at: 7.days.ago..).to_sql
# "SELECT \"clients\".* FROM \"clients\" WHERE \"clients\".\"created_at\" >= '2022-06-24 11:00:50.103960'"
```

Notice the second query uses **>=** instead of **>**. Curiously, rails doesn't seem to be consistent between .. and ...

```ruby
Client.where(created_at: 7.days.ago..).to_sql
# "SELECT \"clients\".* FROM \"clients\" WHERE \"clients\".\"created_at\" >= '2022-06-24 11:06:22.127959'"
Client.where(created_at: 7.days.ago...).to_sql
# "SELECT \"clients\".* FROM \"clients\" WHERE \"clients\".\"created_at\" >= '2022-06-24 11:06:25.876704'"
Client.where(created_at: ..7.days.ago).to_sql
# "SELECT \"clients\".* FROM \"clients\" WHERE \"clients\".\"created_at\" <= '2022-06-24 11:06:32.801216'"
Client.where(created_at: ...7.days.ago).to_sql
# "SELECT \"clients\".* FROM \"clients\" WHERE \"clients\".\"created_at\" < '2022-06-24 11:06:35.632238'"
```

I don't know if this is ActiveRecord's intended behavior, but I avoid using infinite ranges because of this inconsistency.

justindell:
Resolves #321

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1069,10 +1069,16 @@ User.where("created_at > :start AND created_at < end", start: 30.days.ago, end: 
 User.where(created_at: 30.days.ago..7.days.ago)
 
 # bad
-User.where("created_at > ?", 7.days.ago)
+User.where("created_at >= ?", 7.days.ago)
 
 # good
 User.where(created_at: 7.days.ago..)
+
+# note - ranges are inclusive or exclusive of their ending, not beginning
+User.where(created_at: 7.days.ago..)  # produces >=
+User.where(created_at: 7.days.ago...) # also produces >=
+User.where(created_at: ..7.days.ago)  # inclusive: produces <=
+User.where(created_at: ...7.days.ago) # exclusive: produces <
 
 # okish - there is no range syntax that would denote inclusion on one end and
 # exclusion on another.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1069,11 +1069,17 @@ User.where("created_at > :start AND created_at < end", start: 30.days.ago, end:
 User.where(created_at: 30.days.ago..7.days.ago)
```
pirj, line 1069: Should we also change the above bad examples? They are exclusive on both ends.

justindell: Good catch. Updated with b720777

pirj (approved):
Very thorough, thanks a lot!

pirj:
@rubocop/style-guide-editors ping

koic:
Looks good! Thank you!

Issue #155, opened 2015-09-01
`read_attribute(:attr)` and `self[:attr]` not equivalent


amar47shah:
My team is using RuboCop with [Rails cops enabled](https://github.com/bbatsov/rubocop#rails) as a first step in cleaning up a legacy codebase. We found some code like this:

```
  ...
  after_initialize :set_variations

  def set_variations
    if read_attribute(:page_id) && !self.page_id.nil?
      ...
```

Among many others, we got this warning from RuboCop:

```
C: Prefer self[:attr] over read_attribute(:attr).
    if read_attribute(:page_id) && !self.page_id.nil?
       ^^^^^^^^^^^^^^
```

which refers to [this recommendation](https://github.com/bbatsov/rails-style-guide#read-attribute) in the style guide.

And our specs blew up. :( We got several of these:

```
     ActiveModel::MissingAttributeError:
       missing attribute: page_id
     # activerecord-4.0.13/lib/active_record/attribute_methods/read.rb:95:in `block (2 levels) in read_attribute'
     # activerecord-4.0.13/lib/active_record/attribute_methods/read.rb:94:in `fetch'
     # activerecord-4.0.13/lib/active_record/attribute_methods/read.rb:94:in `block in read_attribute'
     # activerecord-4.0.13/lib/active_record/attribute_methods/read.rb:84:in `fetch'
     # activerecord-4.0.13/lib/active_record/attribute_methods/read.rb:84:in `read_attribute'
     # activerecord-4.0.13/lib/active_record/attribute_methods.rb:345:in `[]'
```

It turns out that [`ActiveRecord::AttributeMethods::Read#read_attribute`](http://apidock.com/rails/v4.0.2/ActiveRecord/AttributeMethods/Read/read_attribute) and [`ActiveRecord::AttributeMethods#[]`](http://apidock.com/rails/v4.0.2/ActiveRecord/AttributeMethods/%5B%5D) are not exact synonyms: the latter will raise an `ActiveModel::MissingAttributeError` if the attribute is not found, whereas the former (I think) just returns `nil` and fails silently.

It's true that our example is especially gnarly code (`read_attribute` in an `after_initialize` callback), but I thought I should at least leave a note here that the two methods can't always be swapped safely. Would it be helpful to reword the style recommendation to make that more clear?

bquorning:
I agree, `#[]` *used to be* an alias of `#read_attribute`, but that changed in rails/rails#8056.

[...]

PR #326, opened 2022-08-10
Master

MarcPer:
Addresses https://github.com/rubocop/rails-style-guide/issues/155

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -518,7 +518,7 @@ end
 
 === Read Attribute [[read-attribute]]
 
-Prefer `self[:attribute]` over `read_attribute(:attribute)`.
+Prefer `self[:attribute]` over `read_attribute(:attribute)`. The former raises an `ActiveModel::MissingAttributeError` if the attribute is missing, whereas the latter does not.
 
 [source,ruby]
 ----


pirj (approved):
Thank you!

A small note that to be fair, you can pass a block to `read_attribute` with a handler when the attribute is missing, an this is exactly what [`[]` does](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods.html#method-i-5B-5D).

MarcPer:
@pirj that's a good point, I can adapt the PR to take this into account. How about this?

> Prefer `self[:attribute]` over `read_attribute(:attribute)`, if the latter is used without a block. `self[]` raises an `ActiveModel::MissingAttributeError` if the attribute is missing, whereas `read_attribute` does not.
>
> ```
> [source,ruby]
> ----
> # bad
> def amount
>   read_attribute(:amount) * 100
> end
> 
> # good
> def amount
>   self[:amount] * 100
> end
>
> def amount
>   read_attribute(:amount) { 0 } * 100
> end
> ----
> ```

pirj:
I think this is unnecessary, the PR is good as is.
The point in preferring `[]` is to fail during the development phase making typos in attribute names obvious. I can't think of a real-world usage of a useful `read_attribute` fallback when the attribute doesn't really exist on the model.

MarcPer:
After looking more into the Rails code, I found out the conclusion from the referenced discussion is not exactly correct.

If I just try any random attribute with `[]`, I don't get an exception raised:
```ruby
model[:wrong_attr]
# => nil
```

The `MissingAttributeError` exception comes in when an attribute exists in the model, but is not initialized, as written in the [comments before the method definition](https://github.com/rails/rails/blob/8015c2c2cf5c8718449677570f372ceb01318a32/activerecord/lib/active_record/attribute_methods.rb#L326-L331):
```ruby
person = Person.select('id').first
person[:name]            # => ActiveModel::MissingAttributeError: missing attribute: name
```
The important thing to realize is that the exception is only raised when `Person` has a `name` column. Otherwise the result would be `nil`. Given that, I'd say the text I introduced in this PR is misleading.

PR #345, opened 2023-09-06
Add receiver subsection for Duration Arithmetic

aeroastro:
We prefer an instance of `ActiveSupport::Duration`
as a receiver to calculate relative time like `1.minute.after(created_at)`


README.adoc
```diff
@@ -1860,6 +1860,19 @@ Time.zone.now + 2.days
 2.days.from_now
 ----
 
+Use an instance of `ActiveSupport::Duration` as a receiver to calculate relative time.
+
+[source,ruby]
+----
+# bad - It's confusing to read
+created_at.after(1.minute)
+expiration_date.before(3.days)
```
pirj, line 1869: since/in/ago?
Where are before/after defined?

aeroastro: Thank you for the review, I have fixed the issue at a451ad159c704783060866351a86848dfb5ff8ad.

README.adoc
```diff
@@ -1860,6 +1860,19 @@ Time.zone.now + 2.days
...
+
+[source,ruby]
+----
+# bad - It's confusing to read
+created_at.since(1.minute)
+expiration_date.ago(3.days)
+
+# good
+1.minute.since(created_at)
+3.days.ago(expiration_date)
```
pirj, line 1873: According to a guideline above

> If used with a parameter, prefer `since`, `after`, `until` or `before` instead of `from_now` and `ago`.

`3.days.ago(expiration_date)` would be wrong, too.

And correct options do not map to `ago`/`in`, too.

WDYT of only keeping the `since` example here and in the "bad" section?

pirj (approved):
Thank you!

I don't suggest gathering examples of real-world usages, as the "bad" section examples look terrifyingly bad.
Still, if you have time, here is a good source https://github.com/eliotsykes/real-world-rails.

aeroastro:
Thank you all for the reviews ❤️ 
I have squashed the commits into 1 commit to be merged.

andyw8:
I find to hard to say whether either is 'good' or 'bad' without seeing the examples in a wider context.

aeroastro:
@pirj 

Thank you for the review and mentioning the `2.days.from_now`.

In my opinion, I would like to focus on `1.minute.since(created_at)` for style guide, and also for [rubocop-rails](https://github.com/rubocop/rubocop-rails/pull/1106)

Issue #347, opened 2023-10-13
"Redundant `all`" is too naive


zverok:
I am not sure whether it should be reported as style guide problem, or rubocop-rails problem, but it is a problem :)

Basically, the [styleguide claims](https://rails.rubystyle.guide/#redundant-all) and the cop supports this claim, that those statements are equivalent:
```ruby
User.all.<anymethod>
# and
User.<anymethod>

user.articles.all.<anymethod>
# and
user.articles.<anymethod>
```

The latter is incorrect _at least_ in a case of `delete_all` (which I shoot myself in a foot with just recently, after update to the latest rubocop-rails):
* `user.articles.all.delete_all` is [ActiveRecord::Relation#delete_all](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all) which will just do SQL DELETE, done and forgotten;
* but `user.articles.delete_all` is [ActiveRecord::Associations::CollectionProxy#delete_all](https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete_all) which behaves "according to a deletion strategy", which, say, for `has_many` is by default `:nullify`. 

E.g.:
* `user.articles.all.delete_all` = `DELETE FROM articles WHERE user_id=1` (delete articles in this scope)
* `user.articles.delete_all` = `UPDATE articles WHERE user_id=1 SET user_id=NULL` (remove articles of this scope from the collection "articles of this user")

Which is... Quite a difference. I love Rails and its conventions :heart_eyes: 

So... Probably both the style guide and the cop should be much more cautious here.

koic:
cc @masato-bkn

pirj:
Hey Victor!

Interesting case with delete_all indeed. But how is the styleguide/cop responsible for such an inconsistent and surprising behavior in Rails?
Without any cop or a guideline, wouldn’t you anyway remove the ‘.all’ part from that statement?

I wouldn’t be surprised if it was a many to many association, like ‘peter.friends.delete_all’. But it would remove records from the join table, not nullify, wouldn’t it?

I’d stick to “no .all unless it’s directly called on the model” and let Rails reconsider its default that goes against the principle of least surprise.

Wondering what actions would you guys take to make everyone’s lifes easier?

pirj:
The example ‘person.pets.delete(Pet.find(1))’ looks quite weird. I would probably write it differently, but this notation deserves to exist.

[...]

PR #348, opened 2023-10-19
Add note to redundant-all section about `all` receiver being an association

masato-bkn:
As pointed out https://github.com/rubocop/rails-style-guide/issues/347, when the receiver of `all` is an association, there are methods whose behavior changes when `all` is omitted.

So, I've added a note to caution against omitting these methods.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1103,8 +1103,7 @@ User.where.not('status = ? AND plan = ?', 'active', 'basic')
 
 === Redundant `all` [[redundant-all]]
 
-Using `all` as a receiver for Active Record query methods is redundant. 
-The result won't change without `all`, so it can be removed.
+Using `all` as a receiver is redundant. The result won't change without `all`, so it should be removed.
 
 [source, ruby]
 ----
@@ -1120,6 +1119,17 @@ User.order(:created_at)
 users.where(id: ids)
 user.articles.order(:created_at)
 ----
+
+NOTE: When the receiver for `all` is an association, there are methods whose behavior changes by omitting `all`.
+
+The following methods behave differently without `all`:
+
+* `delete` - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-delete[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete[without all]
+* `delete_all` - https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete_all[without all]
+* `destroy` - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-destroy[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy[without all]
+* `destroy_all` - https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-destroy_all[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy_all[without all]
+
+So, when considering removing `all` from the receiver of these methods, it is recommended to refer to the documentation to understand how the behavior changes.
 
 == Migrations
 


README.adoc
```diff
@@ -1121,6 +1120,20 @@ users.where(id: ids)
 user.articles.order(:created_at)
 ----
 
+NOTE: When the receiver for `all` is an association, there are methods whose behavior changes by omitting `all`.
+
+The following methods behave differently without `all`:
+
+* delete - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-delete[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete[without all]
```
masato-bkn, line 1127: I'm torn about whether to add `delete` and `destroy` to this list for the following reasons:

- these methods are not [the targets for verification of cop](https://github.com/rubocop/rubocop-rails/blob/master/lib/rubocop/cop/rails/redundant_active_record_all_method.rb#L39).
- the expected arguments differ depending on whether `all` is present or not. In the first place, it's not possible to omit `all`.
  - even if omitted, unlike `delete_all`, it seems unlikely that one wouldn't notice the change in behavior:thinking:

README.adoc
```diff
@@ -1121,6 +1120,20 @@ users.where(id: ids)
 user.articles.order(:created_at)
 ----
 
+NOTE: When the receiver for `all` is an association, there are methods whose behavior changes by omitting `all`.
+
+The following methods behave differently without `all`:
```
masato-bkn, line 1125: I think it might be better to leave the specifics of the differences to the official Rails documentation, rather than detailing them in this style guide.

README.adoc
```diff
@@ -1103,8 +1103,7 @@ User.where.not('status = ? AND plan = ?', 'active', 'basic')
 
 === Redundant `all` [[redundant-all]]
 
-Using `all` as a receiver for Active Record query methods is redundant. 
-The result won't change without `all`, so it can be removed.
+When the receiver for Active Record query methods is `all`, the result won't change if `all` is omitted.
```
pirj, line 1106: Can we consider ‘delete*’ “query methods”?
If not, we should emphasize on that.

pirj, line 1106: Can we consider ‘delete*’ “query methods”?
The cop goes beyond just query methods.
What if we briefly mention that ‘all’ is redundant except for those not getting i to details?

masato-bkn: Thank you for your comment!

It's a tough call, but I've always considered `delete*` as query methods. Upon checking the `ActiveRecord::Querying::QUERYING_METHODS` constant in ActiveRecord, it includes methods like `delete*` and `update*`. Given this, in the context of ActiveRecord, it could be argued that `delete*` is also treated as a query method.

However, for those unfamiliar with these specifics, we might consider revising the phrasing.

> What if we briefly mention that ‘all’ is redundant except for those not getting into details?

What specific wording do you think would be appropriate? I'd appreciate any suggestions you might have.

masato-bkn: I made revisions based on the feedback. How does this look? 5efdfa894706dd826a65323decb50e1c6e6956b1

README.adoc
```diff
@@ -1103,8 +1103,7 @@ User.where.not('status = ? AND plan = ?', 'active', 'basic')
 
 === Redundant `all` [[redundant-all]]
 
-Using `all` as a receiver for Active Record query methods is redundant. 
-The result won't change without `all`, so it can be removed.
+Using `all` as a receiver is redundant. The result won't change without `all`, so it could be removed.
```
pirj, line 1106: While we’re here, do you think it’s worth aligning with RFC 2119 and using “should” here instead of a vague “could” (and my prior “can”)?

pirj (approved):
Thanks!

README.adoc
```diff
@@ -1121,6 +1120,20 @@ users.where(id: ids)
...
+
+The following methods behave differently without `all`:
+
+* delete - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-delete[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete[without all]
+
+* delete_all - https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete_all[without all]
+
+* destroy - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-destroy[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy[without all]
+
+* destroy_all - https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-destroy_all[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy_all[without all]
```
koic, line 1133: Please remove the redundant blank lines in the list items and surround the method names with backticks.

```suggestion
* `delete` - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-delete[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete[without all]
* `delete_all` - https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-delete_all[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete_all[without all]
* `destroy` - https://api.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-destroy[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy[without all]
* `destroy_all` - https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-destroy_all[with all] / https://api.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy_all[without all]
```

README.adoc
masato-bkn: Well, 'could' is a bit understated, so I don't think it's a problem to use 'should'.
I've addressed it with https://github.com/rubocop/rails-style-guide/pull/348/commits/9d5d5b145dd717125ec11e9d3dfb61c6ba978b5f.

koic:
This looks good. Can you squash your commits into one?

FYI, in reality, as a reviewer wanting to see the differences, it's sufficient to check using "Compare" link, so there's no need to commits as a work log :-)

README.adoc
masato-bkn: I've addressed it with  https://github.com/rubocop/rails-style-guide/pull/348/commits/d41903b2ecd8b0a9c32e05860dd2253717f5ca8d.

masato-bkn:
Thank you all for your reviews. I have squashed the commits, so please check it.

koic:
Thanks!

PR #355, opened 2024-02-11
Add "Nested Transactions" section

wata727:
See also https://github.com/rubocop/rubocop-rails/pull/1236

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -847,6 +847,28 @@ else
 end
 ----
 
+=== Nested Transactions [[nested-transactions]]
+
+Use `requires_new: true` if you want to avoid nested transactions unintentionally becoming part of the parent transaction.
+`joinable: false` is a private API and should not be used. If used, serious side effects may occur, such as `after_commit` not working properly.
+
+[source,ruby]
+----
+# bad
+ActiveRecord::Base.transaction(requires_new: true, joinable: false) do
+  ActiveRecord::Base.transaction do
+    # ...
+  end
+end
+
+# good
+ActiveRecord::Base.transaction(requires_new: true) do
+  ActiveRecord::Base.transaction(requires_new: true) do
+    # ...
+  end
+end
+----
+
 == Models: Active Record Queries [[activerecord-queries]]
 
 === Avoid Interpolation [[avoid-interpolation]]


pirj:
I didn’t look close yet, but seems interested at a glance https://about.gitlab.com/blog/2021/09/29/why-we-spent-the-last-month-eliminating-postgresql-subtransactions/

wata727:
Thank you for sharing this article. This article seems to be referring to performance issues caused by nested transactions in PostgreSQL.

Please note that this PR is not about whether you should use nested transactions, but only that if you do, you should not use `joinable: false`. This is a fairly partial mention, so I'm honestly not sure if it should be added to the style guide.

pirj (approved):
This makes total sense.

I apologize for an unrelated reference above.

Here are some related references:
 - a PR with a cop corresponding to this guideline https://github.com/rubocop/rubocop-rails/pull/1236
 - nested transactions rspec-rails issue with fruitful discussion https://github.com/rspec/rspec-rails/issues/2598
 - https://github.com/rubocop/rubocop-rails/issues/400 recommends using requires_new: true
 - https://github.com/rubocop/rubocop-rails/pull/414 a corresponding cop (closed as stale)

README.adoc
```diff
@@ -847,6 +847,28 @@ else
 end
 ----
 
+=== Nested Transactions [[nested-transactions]]
```
andyw8, line 850: Similar to what I said in the corresponding RuboCop PR, I think we could group this into a section about avoiding private APIs.

wata727: Thank you for your suggestion.

However, there may be different reasons why private APIs should not be used, so I'm wondering if it's appropriate to group them all in one section.

koic: I think the section should remain as it is because the focus should be more on the issues that arise with nested transactions rather than the use of private APIs.

@wata727 I’ve been reading your blog, and I was wondering if you could elaborate a bit more on `after_commit` in this style guide?

wata727: I'm thinking about this PR again now, and I'm beginning to think that the section might not be suitable as a "style guide." I believe it’s worth having RuboCop find the problems caused by `joinable: false`, but it might be endless to list all such problems in the style guide.

@koic Is there a chance we could add a cop without adding it to the style guide, or do you think problems like this are worth mentioning in the style guide?

PR #356, opened 2024-02-16
Add "Positional arguments" subsection

maxprokopiev:
Related to https://github.com/rubocop/rubocop-rails/pull/1239

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -422,6 +422,8 @@ end
 
 === Enums [[enums]]
 
+=== Hash syntax [[hash-syntax]]
+
 Prefer using the hash syntax for `enum`. Array makes the database values implicit
 & any insertion/removal/rearrangement of values in the middle will most probably
 lead to broken code.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -422,6 +422,8 @@ end
 
 === Enums [[enums]]
 
+=== Hash syntax [[hash-syntax]]
```
pirj, line 425: This should be a sub-section, so four ====

README.adoc
```diff
@@ -440,6 +442,22 @@ class Transaction < ApplicationRecord
 end
 ----
 
+=== Positional arguments [[positional-arguments]]
+
+Prefer using positional arguments instead of keyword arguments when defining enums.
```
pirj, line 447: I’m on the fence if this should be in the guide.
I can see the Rails team trying to get rid of the code that detects the enum attribute name.

I can also see potential mistakes like `enum status: {a: 1}, priority: {high:1,…}`. But is it common?

But how does that stylistically differ for the rest of us? Why do people need to care? If it’s older Rails - use any of the teo. If it’s new, only the positional would work.

maxprokopiev: You're right - since deprecation notice captures the old syntax, seems like we don't need to have a dedicated cop for this at all, I'll close both PRs 😬 

thanks for the feedback!

Issue #328, opened 2022-10-22
Suggestion: Use string literals instead of named routes or URL helpers in tests


ohbarye:
Hi. This is a proposal about named routes usage in tests. I'd like to hear what you think.

### Proposal

Use string literals instead of named routes or URL helpers in tests.

```ruby
# bad
get photos_path
get photo_path(id)
get edit_photo_path(id)
```

```ruby
# good
get "/photos"
get "/photos/#{id}"
get "/photos/#{id}/edit"
```

and so on.

## Reason

Integration test (e.g. request specs, system specs) is to mimic user eyes and actions. A user could see "/photos", but not "photos_path". I agree with the following idea.

>The more your tests resemble the way your software is used, the more confidence they can give you.
>https://twitter.com/kentcdodds/status/977018512689455106

From a programmer's view, 
- the bad pattern slightly increases cognitive load that forces me to build actual URLs in my mind.
- the bad pattern sometimes discourages Test Driven Development.
    - In TDD, a test comes first. At this point, I know actual URL but I don't know the named routes because the `routes.rb` has not been edited yet. (=I don't care about named routes at that moment).

## Possible Objection

named routes possibly have merits in some points.

- You don't have to edit test cases when you edit `routes.rb`. It automatically follows up.
- It could be consistent to use named routes in tests if you use named routes in production code.

## References

- https://guides.rubyonrails.org/routing.html

pirj:
There's some controversy there in the Rails Testing Rails Applications guide.

[Integration](https://guides.rubyonrails.org/testing.html#implementing-an-integration-test):
```ruby
class BlogFlowTest < ActionDispatch::IntegrationTest
  test "can see the welcome page" do
    get "/"
```

[System](https://guides.rubyonrails.org/testing.html#implementing-a-system-test):
```ruby
class ArticlesTest < ApplicationSystemTestCase
  test "viewing the index" do
    visit articles_path
```

Wondering why is it so.

If you don't write route tests and use URL helpers in system/integration, the mapping is not covered. It is risky to have a green CI after changing routes.

RoshanNair22:
I agree with all the responses. But we all have our own way of solving our problems, right?
Yeah, so get on with it.

Also, FYI, if any of our curious readers want to know more about [test driven development](https://thinkpalm.com/blogs/what-is-test-driven-development-and-what-are-its-benefits/), please read the blog written my colluagues.

moznion:
[...]

PR #359, opened 2024-08-22
Add a style guide to suggest to pass a string literal as a URL to http method calls

moznion:
Original discussion: https://github.com/rubocop/rails-style-guide/issues/328

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -231,6 +231,23 @@ match ':controller(/:action(/:id(.:format)))'
 
 Don't use `match` to define any routes unless there is need to map multiple request types among `[:get, :post, :patch, :put, :delete]` to a single action using `:via` option.
 
+=== HTTP URL [[http-url]]
+
+Prefer passing a string literal as a URL to HTTP methods to make the actual URL obvious and to facilitate tracking endpoint path changes.
+
+[source,ruby]
+----
+# bad
+get photos_path
+put photo_path(id)
+post edit_photo_path(id)
+
+# good
+get "/photos"
+put "/photos/#{id}"
+post "/photos/#{id}/edit"
+----
+
 == Controllers
 
 === Skinny Controllers [[skinny-controllers]]


pirj:
I’m in general in favour of this change, but I admit that I’m biased because I’m mostly working on APIs, and the general picture may not be as simple as the limited use case I got used to.

Rails guides [say](https://guides.rubyonrails.org/routing.html) (about the use of path helpers in views):

> `patient_path(@patient)` then the router will generate the path /patients/17. This reduces the brittleness of your view and makes your code easier to understand 

A lot of to digest in this brief note:
1. Reduces brittleness. Path helpers add an abstraction over a detail, the URL path. It’s a design decision whether those paths should be stable or can be bent at will. For the public API or pages that will be bookmarked, shared or referenced from other resources - it makes sense to keep stable. For internal APIs, APIs for own mobile apps or web - not really.
2. Easier to understand. I don’t really take this argument. Pretty sure this can be stretched in a way that the URL will be easier to understand than a path helper.

What is the source of the brittleness?
A mistake causing a change to the URL. Or a deliberate routes refactoring.
It can indeed cause views and redirects to point to the outdated route.
But with tests?

Should we introduce a concept of route stability, and say that this only makes sense as a measure to ensure route stability? And otherwise path helpers are preferrable to reduce brittleness?

What do you think?

moznion:
I appreciate your explanation.

> Easier to understand. I don’t really take this argument. Pretty sure this can be stretched in a way that the URL will be easier to understand than a path helper.

Yeah, I’m on the same page, and the main motivation comes from this.

Anyway, my understanding of the term “brittleness” is related to path-traversal attacks, where malicious “URL-ish” arguments are injected. I completely agree with the approach of using helpers with URL escaping to protect against this type of vulnerability.
However, in testing, this shouldn’t be a problem since it’s just a test. Personally, I believe that other issues, like breaking changes to endpoints, should be caught by the tests.

dvandersluis:
Based on #328 shouldn't this section mention that it applies to tests and not production code?

moznion:
@dvandersluis Thank you for your suggestion; it sounds great to me. However, what would be the best approach: should I include that in the description, or move it under the “Testing” section?

dvandersluis:
Putting it under Testing would make sense to me!

moznion:
@dvandersluis Thank you, I moved it: [1c4cf0e](https://github.com/rubocop/rails-style-guide/pull/359/commits/1c4cf0e5aae086e1e87a1746385828f38a2154f2)

andyw8:
@moznion you might also want to mention tests in the PR title, I was confused the first time reading this.

PR #360, opened 2024-10-21
Un-break PragProg links

arielkirkwood:
It appears PragProg updated "book" to "titles" in their URL structure. Additionally, I added the second edition of the Cucumber book as the destination for "The Cucumber Book".

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1933,10 +1933,10 @@ There are a few excellent resources on Rails style, that you should consider if 
 
 * https://www.informit.com/store/rails-5-way-9780134657677[The Rails 5 Way]
 * https://guides.rubyonrails.org/[Ruby on Rails Guides]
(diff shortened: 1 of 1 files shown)


andyw8 (approved):

andyw8:
I was surprised to see the old URLs didn't redirect and I think there is a bug on ProgProg's site.

If you look at this older snapshot, you can see it redirects with a search:

https://web.archive.org/web/20230423160949/https://pragprog.com/book/rspec3/effective-testing-with-rspec-3

but now it just redirects to a search with a single character.

@noelrappin since you're one of the affected authors, is there someone at ProgProg we could notify about this?

noelrappin:
I'll check with Prag...

Meantime, I should add a link to the Pickaxe book now that we've added an explicit chapter on style...

Issue #361, opened 2024-10-30
Enum: better styling


pirj:
We [suggest](https://github.com/rubocop/rails-style-guide/blob/b077bc2c117880028e055b3d664fa6c69fe6643d/README.adoc?plain=1#L423) the kwargs-style enum definition:
```ruby
  enum type: {
    credit: 0,
    debit: 1
  }
```

This is problematic in two ways.
1. Passing options like `prefix`/`suffix`/`scopes`/`default` is complicated (needs to be [prefixed with an underscore](https://github.com/rails/rails/blob/bc979c5cf4e29e2aba95b7b13ca53e7a02fddee2/activerecord/lib/active_record/enum.rb#L173)) and is not documented
2. Results in weird unrelated errors (mentions the very first defined enum instead of the one that has options)

Ok to change this to use positional args for name and values?
```ruby
  enum :type, {
    credit: 0,
    debit: 1
  }
```
With options:
```ruby
  enum :type, {
    credit: 0,
    debit: 1
  }, suffix: true
```

pirj:
@koic @andyw8 wdyt?

pirj:
Off-topic: Rails 6.1 is [EOL](https://endoflife.date/rails) a month ago, is it a good idea to scrub the guide from legacy content, and leave a note that pre-7 guidelines are only kept in the GitHub repo (in a branch).?

PR #362, opened 2024-12-09
Update `Enum` syntax

crespire:
Updates the enum to show syntax that was introduced in Rails 7.


pirj (approved):
Thank you!

pirj:
@koic do you think if cops are aligned with this new style?

Related: https://github.com/rubocop/rails-style-guide/issues/361

koic:
When a new API is supported only, as with `enum` in this PR, I think it's worth updating the style guide examples accordingly.

Issue #343, opened 2023-08-29
Suggestion: 3-state booleans don't require a default


joerixaop:
I'm not quite sure why it's necessary for booleans to have a default value. Obviously when you are adding a boolean column to an existing table it may be required, because adding non-null columns requires a default in most cases. A newly created table however doesn't have a technical need for a default value for non-null tables, and it seems entirely plausible for a column not to have a good default value (whether it's an integer/string or as in this case a boolean).

pirj:
Have you had a chance to check #275 #337 #43?

joerixaop:
#43 is about where to add a default if you want to add one. I don't want to add one, but I do think a lint about booleans not being nullable is sensible, adding a requirement for a default value in that same lint makes it difficult to use when you do not .

#275 doesn't seem to have any discussion about why requiring a default would be part of the lint that requires a boolean to be not-null. The linked blog post points out what I also pointed out: a default is required when adding a column to an existing table with existing rows, that is guaranteed to be *not* the case when you are adding a boolean column with create_table.

I don't really understand what #337 has to do with default values (boolean or otherwise).

pirj:
If you would decide to send a PR, what would it add/remove/change?

armandmgt:
Jumping in because I recently faced this:
We could enforce the default only when using `add_column` or `change_table`. When creating a new table only the nullability should be checked.

In terms of good/bad examples we could make the following change:
```patch
- # bad - boolean without a default value
+ # bad - boolean without restricted `NULL`
add_column :users, :active, :boolean
+ t.boolean :users, :active, :boolean

- # good - boolean with a default value (`false` or `true`) and with restricted `NULL`
+ # good - boolean with restricted `NULL` and with a default value (`false` or `true`) for existing tables
add_column :users, :active, :boolean, default: true, null: false
add_column :users, :admin, :boolean, default: false, null: false
+ t.boolean :users, :active, :boolean, null: false
```

Would that make sense @pirj ?

pirj:
> We could enforce the default only when using add_column or change_table. When creating a new table only the nullability should be checked.

I didn't get, why "only". Can you please explain, @armandmgt ?

armandmgt:
In the [thoughtbot article](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem) used as reference for this cop, they suggest to add the default because otherwise the migration would fail for existing rows.

[...]

PR #363, opened 2025-01-14
3-state boolean columns emphasis on NOT NULL constraint

armandmgt:
As discussed in https://github.com/rubocop/rails-style-guide/issues/343, the problem of 3-state boolean columns comes from the nullability of the column.

This PR adds emphasis on this. Default values are only required when adding a column to an existing table so `create_table :users { |t| t.boolean :active, null: false }` is completely valid.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1167,19 +1167,30 @@ And you'll have to consider the fact that most non-trivial apps share a database
 
 === 3-state Boolean [[three-state-boolean]]
 
-With SQL databases, if a boolean column is not given a default value, it will have three possible values: `true`, `false` and `NULL`.
+With SQL databases, if a boolean column is nullable, it will have three possible values: `true`, `false` and `NULL`.
 Boolean operators https://en.wikipedia.org/wiki/Three-valued_logic[work in unexpected ways] with `NULL`.
 
 For example in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
 
-To avoid such situations, boolean columns should always have a default value and a `NOT NULL` constraint.
+To avoid such situations, boolean columns should always have a `NOT NULL` constraint.
+
+Note that when adding a boolean column to an existing table, a default value should be put in place. Otherwise the `NOT NULL` constraint will break for existing rows.
 
 [source,ruby]
 ----
-# bad - boolean without a default value
-add_column :users, :active, :boolean
+# bad - boolean column on a new table without a `NOT NULL` constraint
+create_table :users do |t|
+  t.boolean :active
+end
 
-# good - boolean with a default value (`false` or `true`) and with restricted `NULL`
+# bad - adding a boolean without a `NOT NULL` constraint or without a default value
+add_column :users, :active, :boolean
+add_column :users, :active, :boolean, null: false
+
+# good - boolean with a `NOT NULL` constraint, and a default value (`false` or `true`) for existing tables
+create_table :users do |t|
+  t.boolean :active, null: false
+end
 add_column :users, :active, :boolean, default: true, null: false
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1167,19 +1167,30 @@ And you'll have to consider the fact that most non-trivial apps share a database
...
 
 [source,ruby]
 ----
-# bad - boolean without a default value
+# bad - boolean column on a new table without a `NOT NULL` constraint
+create_table :users do |t|
+  t.boolean :active
+end
+
+# bad - adding a boolean without a `NOT NULL` constraint or without a default value
```
pirj, line 1186:
```suggestion
# bad - adding a boolean column without a `NOT NULL` constraint or without a default value to an existing table
```

README.adoc
```diff
@@ -1167,19 +1167,30 @@ And you'll have to consider the fact that most non-trivial apps share a database
...
 add_column :users, :active, :boolean
+add_column :users, :active, :boolean, null: false
 
-# good - boolean with a default value (`false` or `true`) and with restricted `NULL`
+# good - boolean with a `NOT NULL` constraint, and a default value (`false` or `true`) for existing tables
+create_table :users do |t|
+  t.boolean :active, null: false
+end
```
pirj, line 1193:
```suggestion
# good - boolean column with a `NOT NULL` constraint for new tables
create_table :users do |t|
  t.boolean :active, null: false
end

# good - boolean column with a `NOT NULL` constraint, and a default value (`false` or `true`) for existing tables
```

This is fine to not have a default value when creating an new table with a boolean column, right? For those, we explicitly set the value when creating new records, e.g like in a case with the tax_exemption column. Do you have a strong opinion on keeping this as is, or splitting to only enforce the default is fine?

pirj (approved):
Thank you! Minor notes below.

README.adoc
armandmgt: It's definitely clearer with two separate examples 👍 Thanks !

armandmgt:
Hello @koic ! Would you have some time to review this so that we can follow up with the modification of the rubocop cop ?

pirj:
@rubocop/style-guide-editors WDYT?

README.adoc
```diff
@@ -1167,19 +1167,32 @@ And you'll have to consider the fact that most non-trivial apps share a database
 
 === 3-state Boolean [[three-state-boolean]]
 
-With SQL databases, if a boolean column is not given a default value, it will have three possible values: `true`, `false` and `NULL`.
+With SQL databases, if a boolean column is nullable, it will have three possible values: `true`, `false` and `NULL`.
 Boolean operators https://en.wikipedia.org/wiki/Three-valued_logic[work in unexpected ways] with `NULL`.
```
andyw8, line 1171: This article is long and math heavy, is there perhaps a better reference?

README.adoc
```diff
@@ -1167,19 +1167,32 @@ And you'll have to consider the fact that most non-trivial apps share a database
...
 
 For example in SQL queries, `true AND NULL` is `NULL` (not false), `true AND NULL OR false` is `NULL` (not false). This can make SQL queries return unexpected results.
 
-To avoid such situations, boolean columns should always have a default value and a `NOT NULL` constraint.
+To avoid such situations, boolean columns should always have a `NOT NULL` constraint.
+
+Note that when adding a boolean column to an existing table, a default value should be put in place. Otherwise the `NOT NULL` constraint will break for existing rows.
```
andyw8, line 1177: Or it can be added without the constraint, backfilled, and have the constraint added afterwards.

armandmgt: True ! Though adding the default is safer because between the backfill and adding the constraint, new rows might be inserted

Would you like to see a small phrase about the "no constraint, backfill, constraint" procedure ?

PR #369, opened 2025-08-19
Add suggestions for enqueuing many jobs and emails

fatkodima:
Discussion - https://github.com/rubocop/rubocop-rails/issues/1517

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1640,6 +1640,42 @@ Sending emails while generating page response should be avoided.
 It causes delays in loading of the page and request can timeout if multiple email are sent.
 To overcome this emails can be sent in background process with the help of https://github.com/mperham/sidekiq[sidekiq] gem.
 
+=== Enqueuing Many Emails at Once [[enqueuing-many-emails]]
+
+Prefer enqueuing many emails at once instead of one by one in a loop.
+This can greatly reduce the number of round-trips to the queue datastore.
+
+[source,ruby]
+----
+# bad
+users.each { |user| Notifier.welcome(user).deliver_later }
+
+# good
+emails = users.map { |user| Notifier.welcome(user) }
+ActionMailer.deliver_all_later(emails)
+----
+
+NOTE: Rails 8.1 or later is required for using `deliver_all_later`.
+
+== Jobs
+
+=== Enqueuing Many Jobs at Once [[enqueuing-many-jobs]]
+
+Prefer enqueuing many jobs at once instead of one by one in a loop.
+This can greatly reduce the number of round-trips to the queue datastore.
+
+[source,ruby]
+----
+# bad
+ids.each { |id| MyJob.perform_later(id) }
+
+# good
+jobs = ids.map { |id| MyJob.new(id) }
+ActiveJob.perform_all_later(jobs)
+----
+
+NOTE: Rails 7.1 or later is required for using `perform_all_later`.
+
 == Active Support Core Extensions
 
 === `try!` [[try-bang]]


README.adoc
```diff
@@ -1956,6 +1956,30 @@ travel_to(Time.current.to_time)
 freeze_time
 ----
 
+== Performance
```
koic: I felt that the category "Performance" is a bit too broad. I’m not entirely confident, but after looking over the categories, it seems to me that adding "Jobs" under "[Mailers](https://rails.rubystyle.guide/#mailers)" would be a good fit. Strictly speaking, it covers not only Active Job but also Action Mailer, but I made this judgment because the link URL itself includes "jobs".

@rubocop/style-guide-editors If there is a better approach, please suggest it.

fatkodima: Did a split.

Earlopain:
Does this usually say which rails version introduced this? `deliver_all_later` will be part of 8.1 which is not released yet

bbatsov:
I guess it's not a bad idea to add an admonition (e.g. `NOTE:`) to make this clearer.

fatkodima:
Pushed a note.

When implementing a cop, is it ok to create a single cop for both cases (would be a cop that detects single operations in a loop which can be performed in bulk)? I also want it to be extensible to extend with Sidekiq'q `perform_bulk` method.

README.adoc
```diff
@@ -1640,6 +1640,40 @@ Sending emails while generating page response should be avoided.
...
+
+[source,ruby]
+----
+# bad
+users.each { |user| Notifier.welcome(user).deliver_later }
+
+# good
+emails = users.map { |user| Notifier.welcome(user) }
+ActionMailer.deliver_all_later(emails)
+----
```
koic, line 1656: Since it is described as an independent rule, the `NOTE` should also be included here.

README.adoc
```diff
@@ -1640,6 +1640,40 @@ Sending emails while generating page response should be avoided.
 It causes delays in loading of the page and request can timeout if multiple email are sent.
 To overcome this emails can be sent in background process with the help of https://github.com/mperham/sidekiq[sidekiq] gem.
 
+=== Enqueuing Many Emails at Once [[enqueuing-many-emails]]
+
+Prefer enqueuing many emails at once instead of one by one in a loop.
+This can greatly reduce the number of round-trips to the queue datastore.
```
pirj, line 1646: Are you certain that a single batch is better for every store? That multiple messages is worse than one larger message for every protocol?

README.adoc
```diff
@@ -1640,6 +1640,40 @@ Sending emails while generating page response should be avoided.
 It causes delays in loading of the page and request can timeout if multiple email are sent.
 To overcome this emails can be sent in background process with the help of https://github.com/mperham/sidekiq[sidekiq] gem.
 
+=== Enqueuing Many Emails at Once [[enqueuing-many-emails]]
```
pirj, line 1643: There’s https://github.com/toptal/active-job-style-guide, but I don’t know who maintains/owns it now.

README.adoc
fatkodima: Fixed.

README.adoc
```diff
@@ -1640,6 +1640,42 @@ Sending emails while generating page response should be avoided.
...
+Prefer enqueuing many jobs at once instead of one by one in a loop.
+This can greatly reduce the number of round-trips to the queue datastore.
+
+[source,ruby]
+----
+# bad
+ids.each { |id| MyJob.perform_later(id) }
+
+# good
+jobs = ids.map { |id| MyJob.new(id) }
```
pirj, line 1673: I’m puzzled as to why this is needed in the first place. It was always a good practice to accumulate jobs and submit them all at once (to the queue) only when the DB transaction commits, with sidekiq-postpone. Now, there’s a built-in option for that in Sidekiq, and in Rails, too.
Why such explicit manual approach is needed?

README.adoc
fatkodima: Batches are usually in size of 1000, so all the args can possibly be slit into multiple batches.
A single insert is faster than 1000 individual sql queries for database. A single redis command is faster that 1000 redis commands.  Maybe there are cases when multiple commands can be better, but I can't think of such one and in 99.9% cases batches are better.

README.adoc
fatkodima: This is for emails. Another one was added for jobs in a newly created "Jobs" section.

README.adoc
fatkodima: > I’m puzzled as to why this is needed in the first place. It was always a good practice to accumulate jobs and submit them all at once (to the queue) only when the DB transaction commits, with sidekiq-postpone

Haven't ever seen this. It was always done manually where I saw it.

pirj: https://github.com/sidekiq/sidekiq/wiki/Advanced-Options#transactional-push

https://github.com/Envek/after_commit_everywhere

https://edgeguides.rubyonrails.org/active_job_basics.html#transactional-integrity-on-jobs

https://github.com/marshall-lee/sidekiq-postpone

https://github.com/palkan/isolator

fatkodima: But not jobs and emails are enqueued within transactions, most are not. I think the linked gems deal only with transactions.

pirj: Can you give me a few practical examples of code that schedules thousands of jobs/emails and is so trivially simple that it doesn't have to be atomic, and is perfectly linear (wrote to the DB, and just then scheduled thousands of jobs)?

fatkodima: Examples where there is no transactions:
1. Reminding thousands of users about appointments
2. scheduling jobs for thousands of users to sync their calendars
3. schedule jobs for thousands of users to check if they are active
4. notify thousands of course members about the new course item unlocked
5. etc

If I got what was asked.

pirj: Is it really the best practice to schedule thousands of jobs? Or should those better be iterable jobs like https://github.com/sidekiq/sidekiq/wiki/Iteration that you yourself contributed?

Earlopain: I see no big problem with enqueing even millions of jobs if your job storage can handle it. It takes advantage of the job backend to execute them in parallel, depending on the queue size, while iterable jobs would do so in sequence.

BTW, even rails has absorbed a similar feature to what sidekiq has called "Continuations": https://github.com/rails/rails/pull/55127. Or at least I think they're similar, haven't tried it out yet and will only be released in Rails 8.1

That said, for me when I had to enqueue huge amounts of jobs, it was pretty much always a one-off thing so (enqueue) performance didn't really matter that much in those cases.

fatkodima: There are valid use cases for iterable jobs and valid use cases for enqueuing many jobs at once. Not all jobs should be converted to iterable.

I have seen code paths going from minutes to seconds when using batch enqueuing of jobs (and these code paths can run pretty regularly). And when using batches, it will take less time, less requests and load to redis and with less requests a smaller chance that any of these requests fail and the need to start the whole process over.

We discussed examples for a very large number of jobs. But this applies even to hundreds of jobs. The performance can be not as visible for redis, but for database backed queues 1 query vs 100 queries can be a difference (given that it is executed regularly).

Issue #232, opened 2018-08-27
Faster Rails: Use `exists` instead of `present?` to check if a Record Exists


zoras:
Prefer `exists?`instead of `present?` for checking record according to [this blog post](https://semaphoreci.com/blog/2017/03/14/faster-rails-how-to-check-if-a-record-exists.html). 

> ... `exists?`, is even more optimized, and it should be your first choice when checking the existence of a record. It uses the `SELECT 1 ... LIMIT 1` approach, which is very fast.

Also, prefer `any?` and `empty?` when checking for the existence of an association record without any scope. 

> ... `any?` and `empty?` will also produce a very optimized query that uses `SELECT 1 FROM ... LIMIT 1` form, but `any?` will not hit the database again if the records are already loaded into memory.
> 
> This makes `any?` faster by one whole database call when the records are already loaded into memory:

KevinBongart:
Completely agree! Wanna open a PR?

zoras:
Ya sure! Can you guide me where I've to make the changes? 
Didn't know that https://github.com/bbatsov/rubocop & https://github.com/bbatsov/ruby-style-guide are now under https://github.com/rubocop-hq. When did this happen? Any blog posts?

zoras:
Here's another one from a performance [blog post](http://marouenbousnina.com/activerecord-exists-and-blank-under-the-hood/)

> Use blank? if you need to use the records in the application logic.
> Prefer exists? if you care only about the existence check.

KevinBongart:
@zoras I think [the ActiveRecord Queries section](https://github.com/rubocop-hq/rails-style-guide#activerecord-queries) of the README would be a good place to add a bullet point about `exists?` vs. `present?` and `any?` vs. `empty?`.

(The Rubocop repo is more about the tool that can detect rule violations and enforce some of the rules, and the Ruby Style Guide is specific to Ruby whereas the Rails Style Guide includes the ActiveRecord framework.)

3limin4t0r:
Quickly summarizing the methods:

Check for one or more records:

  * `present?` will load all the records and check if there is at least one record present.
  * `exists?` will always query the database with a *COUNT* query to check if at least one record is present.
  *  `any?` behaves like `present?` if records are already loaded and behaves like `exists?` if the records aren't loaded.

Check for no records:

  * `blank?` will load all the records and check if there are no records present.
[...]

PR #371, opened 2025-09-10
Add suggestions for checking if collection is empty

fatkodima:
Reference https://github.com/rubocop/rubocop-rails/issues/1529

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1069,6 +1069,25 @@ User.all.size
 User.all.length
 ----
 
+=== Check if the collection is empty [[check-empty-collection]]
+
+When checking if the unloaded Active Record collection is empty, prefer `any?`/`empty?` over `present?`/`blank?`. The former executes a simple `EXISTS`-like query while the latter loads the whole collection to determine its size.
+
+[source,ruby]
+----
+# bad
+users.where(active: true).present?
+users.where(active: true).blank?
+
+# good
+users.where(active: true).any?
+users.where(active: true).empty?
+
+# good - the collection is used after loading
+users.present?
+users.each(&:notify)
+----
+
 === Where with Ranges [[where-ranges]]
 
 Use ranges instead of defining comparative conditions using a template for scalar values.
--- a/README.adoc
+++ b/README.adoc
@@ -1069,6 +1069,21 @@ User.all.size
 User.all.length
 ----
 
+=== Check if the collection is empty [[check-empty-collection]]
+
+When checking whether an `ActiveRecord::Relation` is empty, prefer `any?`/`empty?` (or `none?`) over `present?`/`blank?`. If the relation is loaded, both variants are equally fast, but if not, the former uses an efficient `EXISTS`/`SELECT 1 … LIMIT 1` query while the latter loads all matching records into memory. (Note: `any?` is only efficient without a block.)
+
+[source,ruby]
+----
+# bad
+user.articles.present?
+user.articles.blank?
+
+# good
+user.articles.any?
+user.articles.empty?
+----
+
 === Where with Ranges [[where-ranges]]
 
 Use ranges instead of defining comparative conditions using a template for scalar values.


README.adoc
```diff
@@ -1069,6 +1069,25 @@ User.all.size
 User.all.length
 ----
 
+=== Check if the collection is empty [[check-empty-collection]]
+
+When checking if the unloaded Active Record collection is empty, prefer `any?`/`empty?` over `present?`/`blank?`. The former executes a simple `EXISTS`-like query while the latter loads the whole collection to determine it's size.
```
andyw8:
```suggestion
When checking if the unloaded Active Record collection is empty, prefer `any?`/`empty?` over `present?`/`blank?`. The former executes a simple `EXISTS`-like query while the latter loads the whole collection to determine its size.
```

README.adoc
```diff
@@ -1069,6 +1069,25 @@ User.all.size
 User.all.length
 ----
 
+=== Check if the collection is empty [[check-empty-collection]]
+
+When checking if the unloaded Active Record collection is empty, prefer `any?`/`empty?` over `present?`/`blank?`. The former executes a simple `EXISTS`-like query while the latter loads the whole collection to determine its size.
```
pirj: [This](https://github.com/rubocop/rails-style-guide/issues/232#issuecomment-425848425) is even more descriptive.
`user.articles.any?` is used just as frequently as `users.where(...).any?`, and in the former case, `any?` behavior is not equivalent to `exists?`, and is not behaving similarly to `empty?`.

README.adoc
```diff
@@ -1069,6 +1069,25 @@ User.all.size
...
+# bad
+users.where(active: true).present?
+users.where(active: true).blank?
+
+# good
+users.where(active: true).any?
+users.where(active: true).empty?
+
+# good - the collection is used after loading
+users.present?
```
pirj: I'm not convinced that preloading the list of records implicitly with `present?` is a practice we should recommend to everyone.

README.adoc
```diff
@@ -1069,6 +1069,25 @@ User.all.size
 User.all.length
 ----
 
+=== Check if the collection is empty [[check-empty-collection]]
```
pirj, line 1072: This is generic. The guideline is specific to "not loaded collection".
It is hard to know if it's loaded or not if it e.g. comes as an argument to a method, or elsewhere:

```ruby
def highlight?
  recent_articles.any?
end
```

README.adoc
pirj: What if there are hundreds of thousands of those records?

README.adoc
```diff
@@ -1069,6 +1069,25 @@ User.all.size
...
+----
+# bad
+users.where(active: true).present?
+users.where(active: true).blank?
+
+# good
+users.where(active: true).any?
+users.where(active: true).empty?
+
+# good - the collection is used after loading
```
pirj: I am referring to the "used" wording above.

Even though the collection will be used a few lines below, maybe only a fraction of it will be actually **used** with a batching iterator like `users.find_each { ...?

Or it will be further queried upon e.g. `users.active`? Or `users.last(3)`.

Or, it will be used in a sub-query, and it doesn't meet the purpose to load them all e.g. `Articles.recent.joins(:author).merge(users)`.

README.adoc
```diff
@@ -1069,6 +1069,25 @@ User.all.size
...
+users.present?
+users.each(&:notify)
```
pirj: https://github.com/rubocop/rails-style-guide/blob/9a1323aa4c4ccfac10ccb9b77b8817e232f8338b/README.adoc#find-each

fatkodima:
@pirj If you have suggestions for better wording/examples, I think it is better to force push into the PR. I used simple examples (not caring about `each` vs `find_each` etc) just for example purposes.

fatkodima:
Updated some wordings accordingly to the suggestions. 
@pirj, @koic please review. 
I want to implement a rubocop-rails cop for this.

README.adoc
```diff
@@ -1069,6 +1069,21 @@ User.all.size
 User.all.length
 ----
 
+=== Check if the collection is empty [[check-empty-collection]]
+
+When checking if the Active Record collection is empty, prefer `any?`/`empty?` over `present?`/`blank?`. If the collection is loaded, both variants are equally fast, but if not, the former executes a simple `EXISTS`-like query while the latter loads the whole collection to determine its size.
```
pirj: Speaking of `empty?` vs `blank?`, there's also a [`none?`](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-none-3F) that is comparable to `empty?`

README.adoc
```diff
@@ -1069,6 +1069,21 @@ User.all.size
...
+
+[source,ruby]
+----
+# bad
+users.where(active: true).present?
+users.where(active: true).blank?
+
+# good
+users.where(active: true).any?
```
pirj: > If the collection is loaded

Won't `users.where(active: true).any?` always execute a `EXISTS` query due to being an AR::Relation? Or will it cache in the SQL cache for the current action?

Practically, loading the whole collection is not feasible anyway, is it?

fatkodima: > Practically, loading the whole collection is not feasible anyway, is it?

Depending of the collection size and how it is used. For example, in this case it is good to load the whole collection
```ruby
users = User.where(...).load
if users.any?
  users.each do ...
```

this will produce 1 query.

Not loading the collection
```ruby
users = User.where(...)
if users.any?
  users.each do ...
```
this will produce 2 queries.

So rubocop can't help here and the developer should identify the problem by himself.

But if the code uses a construction like `where(...).present?` it is always worse than `where(...).any?`.

In this rule description I want to suggest people replacing one by the another.

fatkodima:
Made one more change - maybe it is less confusing now (or more confusing, I don't know 😅)

README.adoc
pirj: Totally agree. There’s a separate guideline for [`for_each`](https://rails.rubystyle.guide/#find-each).

Performing two queries is not such a big deal than accidentally loading the whole, potentially huge dataset in one go.

pirj (approved):
Thank you!

Looks good to me overall.
Personally I would find it easier to understand what the relation is if it was `user.articles` (suggesting there’s an association and the collection is loaded or not) than just `users` (that is harder to imagine not being loaded, like `users = User.where(active: true)`).
But both would be actually correct.

fatkodima:
Thank you for reviews! ❤️ Agreed, converted to `user.articles`.

PR #372, opened 2025-10-11
Prefer JSON's `symbolize_names` option instead of ActiveSupport's `deep_symbolize_keys`

viralpraxis:
ref: https://github.com/rubocop/rubocop-rails/pull/1534

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1727,6 +1727,23 @@ hash.exclude?(:key)
 string.exclude?('substring')
 ----
 
+=== `deep_symbolize_keys` [[deep-symbolize-keys]]
+
+Prefer JSON's `symbolize_names` keyword argument instead of chaining `deep_symbolize_keys`.
+
+Using `symbolize_names: true` is more efficient as it creates symbols during parsing
+rather than requiring a second pass through the data structure.
+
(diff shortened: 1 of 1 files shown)


pirj:
If it’s such a problem, I wonder if AS can somehow mark a freshly parsed object, and issue a warning if deep symbolise is called on it?

pirj:
Cc @byroot

byroot:
> If it’s such a problem

It's wasteful, but not necessarily a problem.

>  I wonder if AS can somehow mark a freshly parsed object

Nah, would mean monkey patching JSON, to then add an instance variable on a Hash object, that's too wasteful and dirty.

viralpraxis:
So does it make sense to implement this rule as a cop?

pirj:
As a cop - certainly yes. As a guideline - I’m uncertain if it’s worth it.

viralpraxis:
Any progress here? it seems to be a tiny, but easy win in terms of performance

> As a cop - certainly yes. As a guideline - I’m uncertain if it’s worth it.

I completely agree with that and I only opened a PR here because it's a mandatory step to merge the PR with the Cop's implementation.

PR #374, opened 2025-10-26
Avoid `Kernel.suppress(Exception)`

viralpraxis:
ref: https://github.com/rubocop/rubocop-rails/issues/1547

https://api.rubyonrails.org/classes/Kernel.html#method-i-suppress

Using `Kernel.suppress(Exception) { ?? }` is effectively equivalent to:

```ruby
begin
  ??
rescue Exception
end
```

Since rescuing `Exception` directly is discouraged by `Lint/RescueException`, it makes sense to discourage this usage as well.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1768,6 +1768,23 @@ datetime.to_fs(:db)
 42.to_fs(:human)
 ----
 
+=== Do not rescue `Exception` via `Kernel.suppress` [[kernel-suppress-exception]
+
+Using `Kernel.suppress(Exception)` is effectively equivalent to rescuing `Exception`, which is discouraged.
+
+[source,ruby]
+----
+# bad
+suppress(Exception) do
+  do_something
+end
+
+# good
+supress(StandardError) do
+  do_something
+end
+----
+
 == Time
 
 === Time Zone Config [[tz-config]]


README.adoc
```diff
@@ -1768,6 +1768,27 @@ datetime.to_fs(:db)
...
+end
+
+# good
+begin
+  do_something
+rescue Exception => e
+  process_exception(e)
+
+  raise
+end
```
koic, line 1785: I'm not sure why this is considered a good case. Considering https://rubystyle.guide/#no-blind-rescues and the [implementation of Active Support's `suppress`](https://github.com/rails/rails/blob/v8.1.0/activesupport/lib/active_support/core_ext/kernel/reporting.rb#L33-L44), the real issue seems to be that it rescues `Exception`, right?
So, a better good example would probably use something like `suppress(StandardError)` instead.

It might also be good to mention [https://rubystyle.guide/#no-blind-rescues](https://rubystyle.guide/#no-blind-rescues).

viralpraxis: I thought that the good example here is fine since it seems like [it will be allowed soon](https://github.com/rubocop/rubocop/pull/14579).

Anyways, perhaps `StandardError` would be better here

koic: > I thought that the good example here is fine since it seems like [it will be allowed soon](https://github.com/rubocop/rubocop/pull/14579).

That case shows re-raising, but this example doesn't seem to show that way.

viralpraxis: Not sure I'm following you here.

The initial example was

```ruby
begin
  do_something
rescue Exception => e
  process_exception(e)

  raise
end
```

which shows re-raising as well. What am I missing?

koic: I understand https://rubystyle.guide/#no-blind-rescues and this rule as aiming to ensure that exceptions are subclasses of `StandardError`, rather than rescuing the broad `Exception` class. The use of `Exception` in a re-raise is an exceptional case where it can be allowed in https://github.com/rubocop/rubocop/pull/14579, but that's not the main focus. As for `begin`...`rescue`, that’s already covered by the Ruby Style Guide's rule, while this rule seems intended to focus on the Rails API method `suppress`.
In other words, I think it works as a minimal example in its current form.

koic:
One concern is that https://rubystyle.guide/#no-blind-rescues is a rule intended to prevent catching exceptions with incorrect handling. On the other hand, `suppress` is a method intended to suppress exception handling.
In other words, the goals are different, so I'm not fully convinced that `suppress(Exception)` should truly be treated as a bad case.

viralpraxis:
Are there any legitimate uses of `suppress(Exception)` in the context of Rails? I can’t think of any.

On the other hand, its easy to use `Exception` instead of `StandardError` out of inexperience just as with the "regular rescue" syntax.

Anyway, I don't insist of making this a rule, but I think a cop would be beneficial

PR #377, opened 2026-04-23
Add Avoid `in_batches` without `disable_ddl_transaction`

bart-westenenk-bex:
Having your migration do a data migration with `in_batches` is good. However, if the default DDL transaction is still active, it will still commit all work in one go to the database and `in_batches` is useless.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1384,6 +1384,52 @@ class DropUsers < ActiveRecord::Migration
 end
 ----
 
+=== Avoid `in_batches` without `disable_ddl_transaction!` [[avoid-in-batches-without-disable-ddl-transaction]]
+
+When using `in_batches` in a migration, make sure to disable the DDL transaction by adding `disable_ddl_transaction!` to the migration class.
+If you don't disable the DDL transaction, the default DDL transaction will wrap the entire migration.
+This means `in_batches` will not commit work in smaller units and instead locks all touched records in the database in the same transaction.
+This leads to long locks and downtime on production databases.
+
+[source,ruby]
+----
+# bad - missing disable_ddl_transaction!
+class BackfillUsers < ActiveRecord::Migration
+  def up
+    User.in_batches do |batch|
+      batch.update_all(status: "active")
+    end
+  end
+end
+
+# good - in_batches commits work in smaller units of work
+class BackfillUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def up
+    User.in_batches do |batch|
+      batch.update_all(status: "active")
+    end
+  end
+end
+
+# bad - missing disable_ddl_transaction!
+class BackfillUsers < ActiveRecord::Migration
+  def up
+    User.in_batches.update_all(status: "active")
+  end
+end
+
+# good - in_batches commits work in smaller units of work
+class BackfillUsers < ActiveRecord::Migration
+  disable_ddl_transaction!
+
+  def up
+    User.in_batches.update_all(status: "active")
+  end
+end
+----
+
 == Views
 
 === No Direct Model View [[no-direct-model-view]]


bart-westenenk-bex:
@bbatsov Hi, would you be able to have a look at this?

pirj:
Isn’t disabling the ddl transaction a rule of thumb for any data migrations? https://github.com/ankane/strong_migrations#good-7

bart-westenenk-bex:
Yes but no. For some data migrations that you want to be atomic, you may want to keep the transaction so when it fails you do not have some in between state. You may argue that in that you need to make the transaction around the update explicit though, but that isn't really Railsy.

pirj:
The original ["outside of a transaction"](https://github.com/ankane/strong_migrations/commit/44412fe1ee8671306ec2f276f0898a461cfb950f) part stems from [this article](https://wework.github.io/data/2015/11/05/add-columns-with-default-values-to-large-tables-in-rails-postgres/), which assumes two things:
 - adding a row
 - backfilling it in the same transaction that adds the row
But nevertheless, this recommendation still stands in its generic form for those who follow the zero-downtime migrations practices:
> Use ... a separate migration with `disable_ddl_transaction!` [for backfilling in a migration]

You suggest adding a softer wording, but to a wider audience:
> When using `in_batches` in a migration, make sure to disable the DDL transaction

I clearly see the problem you're bringing up.


> data migrations that you want to be atomic

"Atomic" and `in_batches` don't mix well.


> need to make the transaction around the update explicit though, but that isn't really Railsy

Agree, not Railsy. And I wouldn't suggest that, as it splits the purpose of `in_batches` into:
1) fetching records in batches
2) performing something on those records in a transaction
and that makes it more confusing than it should be.


It depends on what you're doing exactly inside the block passed to `in_batches`: update all records similarly, update each record separately, perform external API calls, a combination of those, ...
Since it's case-dependent, do you feel that there should be a single guideline to define how it should be done in all cases?

bart-westenenk-bex:
I don't think it would be a good idea either way to do API calls in a migration, that is better suited for some background jobs. But maybe I am missing a legitimate usecase. So I am not sure what other cases we have to consider here.

pirj:
Three cases just from [`in_batches` docs](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-in_batches):
1. Operation on all at once (`update_all`)- no transaction, single statement
2. [`each`](https://api.rubyonrails.org/classes/ActiveRecord/Batches/BatchEnumerator.html#method-i-each_record)/`each_record` that can trigger individual updates, non-atomically
3. `each_record` with a block that also causes cascading updates via callbacks (`Person.in_batches.each_record(&:party_all_night!)`)

There’s no guarantee that API calls won’t happen inside those callbacks, even if not explicitly made from the migration itself.

bart-westenenk-bex:
I am not really sure if you are arguing in favor or against the proposed guideline, so I'll just post a clarification of my points hoping that clears up things.

My goal was to make sure `in_batches` is only used if it actually makes the migration more performant. Afaik, having the `in_batches` wrapped in a transaction will _always_ negate the effects of using `in_batches` (ie. reducing the amount of row locks). Regardless of what is done with the `in_batches` enumerator later on.
In both read only and updating scenarios, the row locks will be an issue.
We have had multiple issues during a deployment because someone wrote a 'performant' migration that kept transactions in mind, but forgot to add a `disable_ddl_transaction!` and took production down with it.

> > data migrations that you want to be atomic
>
> "Atomic" and in_batches don't mix well.

What I meant with this bit is that you may want to have the migration atomic _per batch_. In the case the migration fails midway, you can continue the migration from the latest batch (assuming you have a where clause that can detect which part of the data still needs to be migrated).

Note that I am not saying anything about making each _batch_ atomic for this best practice. It is in my opinion still a best practice, but different to this one and would need it's own PR.

> There’s no guarantee that API calls won’t happen inside those callbacks, even if not explicitly made from the migration itself.

Also, I think it would be better to defer API calls to after the transaction has been committed, since otherwise you may send an external party that a record has been created ... but it couldn't be created due to some other conflict in the database. Causing the external party to have mismatched data. But again, different best practice :)

bart-westenenk-bex:
@pirj Is there anything else that you think is blocking to get this merged?

PR #378, opened 2026-05-13
Add no-include-view-helpers rule to Views section

mathieugoedhart96:
Adds a new style guide entry in the Views section discouraging the use of
  `include` with view helper modules (e.g. `UserHelper`, `ApplicationHelper`,
  `ActionView::Helpers::NumberHelper`) outside of view contexts such as controllers, models, and jobs.

View helpers depend on the view context and including them elsewhere may cause methods to fail silently or raise errors.

There are two alternatives:
  - The `helpers` proxy for controllers
  - Extracting shared behavior into a concern under `app/concerns`

This corresponds to the `Rails/IncludeViewHelper` cop being proposed in rubocop/rubocop-rails#1609.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1420,6 +1420,69 @@ In these cases, an undefined instance variable will not raise an exception where
 <%= course.description %>
 ----
 
+=== No View Helpers Outside Views [[no-include-view-helpers]]
+
+Do not `include` view helper modules (`ActionView::Helpers`) outside of the view context.
+Doing so blurs the boundaries between layers of the stack, leading to tangled design—for example, business logic creeping into helpers.
+It also introduces technical risks: view helpers depend on the view context, so including them in controllers, models, or jobs may cause methods to fail silently or raise errors.
+
+Modules in `app/helpers` are automatically included into all views by Rails.
+If shared behavior is needed across multiple classes, extract it into a concern under `app/concerns` or delegate to a plain Ruby object instead.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1420,6 +1420,54 @@ In these cases, an undefined instance variable will not raise an exception where
 <%= course.description %>
 ----
 
+=== No View Helpers Outside Views [[no-include-view-helpers]]
+
+Do not `include` view helper modules (`ActionView::Helpers`) outside of the view context.
+View helpers depend on the view context; including them in controllers, models, or jobs may cause methods to fail silently or raise errors because the view context is absent.
```
andyw8: I agree with the principle, but I'd say the primary problem is that it blurs the boundaries between different levels of the stack, leading to a tangled design, e.g. business logic being added to helper.

README.adoc
```diff
@@ -1420,6 +1420,54 @@ In these cases, an undefined instance variable will not raise an exception where
...
+
+Modules in `app/helpers` are automatically included into all views by Rails.
+If shared behavior is needed across multiple classes, extract it into a concern under `app/concerns` instead.
```
andyw8: There are other options too, e.g. delegating to a PORO.

README.adoc
mathieugoedhart96: Good point. I’ve added it to the text.

mathieugoedhart96:
@andyw8 I’ve incorporated your feedback

PR #383, opened 2026-09-16
Fix examples using outdated Rails APIs

vivekmiyani:
This PR updates three examples in the guide that use outdated Rails methods:

- `attr_accessible`: Removed in Rails 4
- `enum` (keyword arguments): Removed in Rails 8.0. This fixes a missed occurence from #362 in the Macro Style Methods section
- `update_attributes`: Removed in Rails 6.1 and replaced with `update`


pirj (approved):
Thank you 🤩

